PR #1, opened 2023-05-27
Initial commit

LinZhihao-723:
This first PR constructs the project structure of clp-ffi-py. It structures the src files and sets up necessary build tools. Currently, only 4-byte encoding methods are implemented (which is used by clp-loglib-py). We can incrementally port other methods later.


README.md
```diff
@@ -1,2 +1,58 @@
 # clp-ffi-py
-clp-ffi-py is a Python library to encode log messages with CLP, and work with the encoded messages using a foreign function interface (FFI).
+
+This module provides Python packages to interface with [CLP Core Features](https://github.com/y-scope/clp/tree/main/components/core) through CLP's FFI (foreign function interface). Currently, this library provides native methods to encode/decode log messages with [CLP](https://github.com/y-scope/clp).
```
davidlion: Shorten long lines by using semantic line feeds (https://rhodesmill.org/brandon/2012/one-sentence-per-line/) and references (https://www.markdownguide.org/basic-syntax#reference-style-links)

davidlion (changes requested):

README.md
```diff
@@ -1,2 +1,58 @@
...
+
+## Compatibility
+
+Tested on Python 3.6 and 3.8 and should work on any newer version.
+
+## Building/Packaging
+
+To maually build a package for distribution, run the following step. It should generate both .tar.gz pacakge and .whl package under `./dist/`.
```
kirkrodrigues: * maually -> manually
* pacakge -> package

README.md
```diff
@@ -1,2 +1,58 @@
...
+
+1. Create and enter a virtual environment:
+   `python3.8 -m venv venv; . ./venv/bin/activate`
```
kirkrodrigues: * `python3.8` -> `python3`?
* If we want to do these commands in one line, it should really be `&&` instead of `;`, right?

README.md
kirkrodrigues, line 1: Personally, I prefer putting commands that users need to copy in a code **block** rather than an **inline** code snippet.
E.g.

Code block:

```shell
python3 -m venv venv
```

Inline code: `python3 -m venv venv`

It's faster to copy the former since you just need to copy the entire line. Moreover, GitHub's Markdown renderer creates a copy button for code blocks making it even easier.

README.md
```diff
@@ -1,2 +1,58 @@
...
+2. Install development dependencies:
+   `pip install -r requirements-dev.txt`
```
kirkrodrigues: Can we use dev dependencies in the toml rather than a requirements.txt file?

src/clp_ffi/ErrorMessage.hpp
```diff
@@ -0,0 +1,15 @@
+#ifndef CLP_PY_ERROR_MESSAGE
```
kirkrodrigues:
```suggestion
#ifndef CLP_FFI_PY_ERROR_MESSAGE
```

src/clp_ffi/ErrorMessage.hpp
```diff
@@ -0,0 +1,15 @@
...
+#define CLP_PY_ERROR_MESSAGE
+
+namespace clp_ffi_py::ErrorMessage {
+constexpr char arg_parsing_error[] = "Native preamble encoder failed to parse Python arguments.";
+constexpr char out_of_memory_error[] = "Failed to allocate memory.";
+} // namespace clp_ffi_py::ErrorMessage
+
+namespace clp_ffi_py::ErrorMessage::Encoding {
```
kirkrodrigues: Can we move this nested namespace into the previous namespace?

src/clp_ffi/clp_four_byte_encoder.cpp
```diff
@@ -0,0 +1,40 @@
+#include "encoding_method.hpp"
+
+/**
+ * Method table
+ */
+static PyMethodDef EncoderMethods[] = {
+        {"encode_preamble",
+         clp_ffi_py::encoder::four_byte_encoding::encode_preamble,
+         METH_VARARGS,
+         "encode the preamble with a reference timestamp, a timestamp format, and a time zone."},
```
kirkrodrigues: I was reading a bit about how these work and this is actually where we should be putting the docstrings that are currently in `CLPFourByteEncoder.pyi`. There are two caveats:
* We need to specify the method's signature at the top of the method followed by `--\n\n`. See [here](https://stackoverflow.com/questions/1104823/python-c-extension-method-signatures-for-documentation) for more details.
* We should use the [`PyDoc_STRVAR`](https://docs.python.org/3/c-api/intro.html#c.PyDoc_STRVAR) macro.

src/clp_ffi/clp_four_byte_encoder.cpp
```diff
@@ -0,0 +1,40 @@
...
+         clp_ffi_py::encoder::four_byte_encoding::encode_message,
+         METH_VARARGS,
+         "encode the log message content"},
+
+        {"encode_timestamp_delta",
+         clp_ffi_py::encoder::four_byte_encoding::encode_timestamp_delta,
+         METH_VARARGS,
+         "encode the timestamp delta."},
+
+        {NULL, NULL, 0, NULL}};
```
kirkrodrigues: * Can we use `nullptr` instead or can we rely on [zero initialization](https://en.cppreference.com/w/cpp/language/zero_initialization)?
* Given the length of this `PyMethodDef` block, I think it's clearer to have the final `}` on its own line so that it's easy to see where this array ends.

src/clp_ffi/encoding_method.hpp
```diff
@@ -0,0 +1,46 @@
+#ifndef CLP_PY_ENCODING_METHOD
+#define CLP_PY_ENCODING_METHOD
+
+#include <Python.h>
```
kirkrodrigues: * Should `#define PY_SSIZE_T_CLEAN` on the previous line. See https://docs.python.org/3/c-api/arg.html#strings-and-buffers
* Should add a comment that Python.h must be included before any other includes as noted in https://docs.python.org/3/c-api/intro.html#include-files

src/clp_ffi/encoding_method.cpp
```diff
@@ -0,0 +1,121 @@
+#include <chrono>
+
+#include "../clp/components/core/src/ffi/encoding_methods.hpp"
+#include "../clp/components/core/src/ffi/ir_stream/encoding_methods.hpp"
+
+#include "ErrorMessage.hpp"
+#include "encoding_method.hpp"
```
kirkrodrigues: Should be the first include because:

* That's what the style guide dictates.
* That will cause `Python.h` to be included before any other includes.

src/clp_ffi/encoding_method.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+namespace clp_ffi_py::encoder::four_byte_encoding {
+/**
+ * Given the reference timestamp, the timestamp format, and the time zone,
```
kirkrodrigues: * In each of these methods, could we omit mentioning the arguments in the method's description since we mention them in the description for `args`?
* Use command form for method descriptions.

src/clp_ffi/encoding_method.hpp
```diff
@@ -0,0 +1,46 @@
...
+ * encode the preamble
+ * @param self
+ * @param args, the reference timestamp, the timestamp format, and the time
```
kirkrodrigues: In each of these methods, don't need a comma after `args`.

src/clp_ffi/encoding_method.hpp
```diff
@@ -0,0 +1,46 @@
...
+ * zone are expected in sequence
+ * @return python bytearray that contains the encoded preamble
```
kirkrodrigues: Nits:
```suggestion
 * @return A Python byte array containing the encoded preamble
```

src/clp_ffi/encoding_method.hpp
```diff
@@ -0,0 +1,46 @@
...
+ */
+PyObject* encode_preamble (PyObject* self, PyObject* args);
+
+/**
+ * Given the timestamp delta of the last log message and the content of the
+ * current log message, encode the message using 4-byte encoding
+ * @param self
+ * @param args, timestamp delta and input byte buffer of the current log
+ * message are expected in sequence
+ * @return python bytearray that contains the encoded message
```
kirkrodrigues: It also contains the timestamp delta, right?

src/clp_ffi/encoding_method.hpp
```diff
@@ -0,0 +1,46 @@
...
+ */
+PyObject* encode_message (PyObject* self, PyObject* args);
+
+/**
+ * Encode the given timestamp delta into IR and write it into the buffer
+ * using 4-byte encoding
+ * @param self
+ * @param args, timestamp delta
+ * @return python bytearray that contains the encoded message
```
kirkrodrigues: Copy paste error?

src/clp_ffi/encoding_method.hpp
```diff
@@ -0,0 +1,46 @@
+#ifndef CLP_PY_ENCODING_METHOD
+#define CLP_PY_ENCODING_METHOD
+
+#include <Python.h>
+
+namespace clp_ffi_py::encoder::four_byte_encoding {
```
kirkrodrigues: Style guide dictates that each namespace gets its own directory, so this should be in an `encoder` directory.

src/clp_ffi/encoding_method.hpp
```diff
@@ -0,0 +1,46 @@
+#ifndef CLP_PY_ENCODING_METHOD
```
kirkrodrigues:
```suggestion
#ifndef CLP_FFI_PY_ENCODING_METHOD
```

LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2023 YScope Inc.
```
kirkrodrigues: If you read from line 178, this paragraph is describing how the Apache License should be applied to a piece of work. E.g., if you wanted to apply the Apache license to a file, you wouldn't paste this entire file into the header of your file; instead, you would copy from line 189 to 201 and then modify the boilerplate values. For an example, see [here](https://github.com/openstack/nova/blob/master/nova/cache_utils.py).

So it's incorrect to change line 189 from the boilerplate value that it had.

src/clp_ffi/encoding_method.hpp
kirkrodrigues, line 1: Filename should be `encoding_methods` (plural) right?

src/clp_ffi/encoding_method.cpp
```diff
@@ -0,0 +1,121 @@
+#include <chrono>
```
kirkrodrigues: Necessary?

src/clp_ffi/ErrorMessage.hpp
```diff
@@ -0,0 +1,15 @@
+#ifndef CLP_PY_ERROR_MESSAGE
+#define CLP_PY_ERROR_MESSAGE
+
+namespace clp_ffi_py::ErrorMessage {
```
kirkrodrigues: * Technically namespaces should be named with snake case.
* It might be more intuitive to call this `error_messages` (plural)

.clang-format
```diff
@@ -0,0 +1,29 @@
...
+---
+Language: Cpp
+AccessModifierOffset: -4
+AllowShortFunctionsOnASingleLine: Inline
+AllowShortIfStatementsOnASingleLine: Never
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakTemplateDeclarations: Yes
+BinPackArguments: false
+BinPackParameters: false
+ContinuationIndentWidth: 8
```
kirkrodrigues, line 17: Let's add `InsertNewlineAtEOF: true`

.clang-format
```diff
@@ -0,0 +1,29 @@
+---
+BasedOnStyle: LLVM
+ColumnLimit: 100
+IndentWidth: 4
+---
+Language: Cpp
+AccessModifierOffset: -4
```
kirkrodrigues, line 7: Let's add `AlignAfterOpenBracket: AlwaysBreak`

src/clp_ffi/encoding_method.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+#include "../clp/components/core/src/ffi/encoding_methods.hpp"
+#include "../clp/components/core/src/ffi/ir_stream/encoding_methods.hpp"
+
+#include "ErrorMessage.hpp"
+#include "encoding_method.hpp"
+
+namespace clp_ffi_py::encoder::four_byte_encoding {
+PyObject* encode_preamble (PyObject* self, PyObject* args) {
```
kirkrodrigues: Can wrap unused parameters with `Py_UNUSED`.

src/clp_ffi/encoding_method.cpp
```diff
@@ -0,0 +1,121 @@
...
+    ffi::epoch_time_ms_t ref_timestamp;
+    char const* input_timestamp_format;
+    char const* input_timezone;
+
+    if (false ==
+        PyArg_ParseTuple(args, "Lss", &ref_timestamp, &input_timestamp_format, &input_timezone)) {
```
kirkrodrigues: Can we use `s#` for the strings? Similarly, we can use the `#` variant in the other methods.

src/clp_ffi/encoding_method.cpp
```diff
@@ -0,0 +1,121 @@
...
+        PyErr_SetString(PyExc_RuntimeError, clp_ffi_py::ErrorMessage::arg_parsing_error);
```
kirkrodrigues: The [docs](https://docs.python.org/3/c-api/intro.html#exceptions) suggest that:
* the API methods will throw a Python exception on failure and we should not throw our own exception since that will overwrite the already thrown exception.
* on exception, we should return `NULL` (`nullptr`) or `-1` depending on the method's return type.

src/clp_ffi/encoding_method.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+    const std::string timestamp_format(input_timestamp_format);
+    const std::string timezone(input_timezone);
+    const std::string timestamp_pattern_syntax = "";
+    std::vector<int8_t> ir_buf;
+
+    if (false ==
+        ffi::ir_stream::four_byte_encoding::encode_preamble(
+                timestamp_format, timestamp_pattern_syntax, timezone, ref_timestamp, ir_buf)) {
+        PyErr_SetString(PyExc_NotImplementedError,
```
kirkrodrigues: Why `PyExc_NotImplementedError` instead of something like `PyExc_ValueError`?

src/clp_ffi/encoding_method.cpp
```diff
@@ -0,0 +1,121 @@
...
+                        clp_ffi_py::ErrorMessage::Encoding::preamble_error);
+        Py_RETURN_NONE;
+    }
+
+    PyObject* output_bytearray =
+            PyByteArray_FromStringAndSize(reinterpret_cast<char*>(ir_buf.data()), ir_buf.size());
```
kirkrodrigues: Use clp's `size_checked_pointer_cast`

src/clp_ffi/encoding_method.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+    if (false ==
+        ffi::ir_stream::four_byte_encoding::encode_preamble(
+                timestamp_format, timestamp_pattern_syntax, timezone, ref_timestamp, ir_buf)) {
+        PyErr_SetString(PyExc_NotImplementedError,
+                        clp_ffi_py::ErrorMessage::Encoding::preamble_error);
+        Py_RETURN_NONE;
+    }
+
+    PyObject* output_bytearray =
```
kirkrodrigues: Nit: Use `auto` or you could return the function's return value directly

src/clp_ffi/encoding_method.cpp
```diff
@@ -0,0 +1,121 @@
...
+    ffi::epoch_time_ms_t delta;
+    Py_buffer input_buffer;
+    if (false == PyArg_ParseTuple(args, "Ly*", &delta, &input_buffer)) {
+        PyErr_SetString(PyExc_RuntimeError, clp_ffi_py::ErrorMessage::arg_parsing_error);
+        Py_RETURN_NONE;
+    }
+
+    std::string logtype;
+    std::vector<int8_t> ir_buf;
+    ir_buf.reserve(input_buffer.len * 2);
```
kirkrodrigues: Can we add a comment explaining why we do this?

src/clp_ffi/encoding_method.cpp
```diff
@@ -0,0 +1,121 @@
...
+    std::string_view msg(static_cast<char const*>(input_buffer.buf), input_buffer.len);
+
+    // Encode the message
+    if (false == ffi::ir_stream::four_byte_encoding::encode_message(msg, logtype, ir_buf)) {
+        PyBuffer_Release(&input_buffer);
```
kirkrodrigues: If we decide to stick with `y*`, it's probably beneficial to use a `unique_ptr` with a custom deleter like what we do in [`clp-ffi-java`](https://github.com/y-scope/clp-ffi-java/blob/main/src/main/cpp/libclp_ffi_java/JavaPrimitiveArrayElementsDeleter.cpp).

src/clp_ffi/clp_four_byte_encoder.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Module definition
+ */
+static struct PyModuleDef clp_four_byte_encoder = {
+        PyModuleDef_HEAD_INIT, "CLPFourByteEncoder", NULL, -1, EncoderMethods};
```
kirkrodrigues: * Let's define a module docstring.
* `-1` means the module has global state, right? So should we set it to `0` instead?

src/clp_ffi/clp_four_byte_encoder.cpp
```diff
@@ -0,0 +1,40 @@
...
+         clp_ffi_py::encoder::four_byte_encoding::encode_timestamp_delta,
+         METH_VARARGS,
+         "encode the timestamp delta."},
+
+        {NULL, NULL, 0, NULL}};
+
+/**
+ * Module definition
+ */
+static struct PyModuleDef clp_four_byte_encoder = {
```
kirkrodrigues: Don't need the `struct`

src/clp_ffi/clp_four_byte_encoder.cpp
```diff
@@ -0,0 +1,40 @@
...
+        PyModuleDef_HEAD_INIT, "CLPFourByteEncoder", NULL, -1, EncoderMethods};
+
+/**
+ * Module initialization
+ */
+PyMODINIT_FUNC PyInit_CLPFourByteEncoder (void) {
```
kirkrodrigues: Do we need the `void`?

src/clp_ffi/ErrorMessage.hpp
```diff
@@ -0,0 +1,15 @@
...
+constexpr char arg_parsing_error[] = "Native preamble encoder failed to parse Python arguments.";
+constexpr char out_of_memory_error[] = "Failed to allocate memory.";
```
kirkrodrigues: Unused?

src/clp_ffi/ErrorMessage.hpp
```diff
@@ -0,0 +1,15 @@
...
+} // namespace clp_ffi_py::ErrorMessage
+
+namespace clp_ffi_py::ErrorMessage::Encoding {
+constexpr char timestamp_error[] = "Timestamp delta > signed int32 currently unsupported";
```
kirkrodrigues: Using specific errors based on the IR-encoding method implementations does put a burden on this library to constantly stay up-to-date with the implementations of the IR encoding methods. A better option is probably to use exceptions within the IR encoding methods and return error strings that way. So personally, I would keep these errors generic for now until such time someone improves the IR encoding methods (an error with no details is better than an error with incorrect details).

README.md
```diff
@@ -1,2 +1,58 @@
...
+3. Pull all submodules in preparation for building:
+   `git submodule update --init --recursive`
+4. Build:
+   `python -m build`
+
+## Testing
+
+1. Create and enter a virtual environment:
+   `python -m venv venv; .  ./venv/bin/activate`
```
kirkrodrigues: Nit: Extra space after `.`

setup.py
```diff
@@ -0,0 +1,31 @@
...
+    ],
+    define_macros=[
+        ("SOURCE_PATH_SIZE", str(len(os.path.abspath("./src/clp/components/core"))))
+    ]
+)
+
+
+setup(
+    name="clp_ffi_py",
+    version="0.0",
```
kirkrodrigues: Seems redundant to specify the version and description here (since it's already in the toml). Can we get rid of it?

kirkrodrigues (changes requested):
* Let's add the `.clang-tidy` config from log-surgeon.
* I haven't reviewed the "packaging" (GitHub action, README, etc.) files very carefully. I will when we're closer to finishing the PR.

src/clp_ffi/clp_four_byte_encoder.cpp
LinZhihao-723: 1. I would prefer using {nullptr, nullptr, 0, nullptr} to indicate the end of the method table explicitly.
2. the position of "}" is auto formatted by clang-format

src/clp_ffi/encoding_method.hpp
LinZhihao-723: Yes, added

src/clp_ffi/encoding_method.hpp
LinZhihao-723: Solved

src/clp_ffi/encoding_method.cpp
LinZhihao-723: Removed

src/clp_ffi/encoding_method.cpp
LinZhihao-723: I will remove the error string and return nullptr instead.

src/clp_ffi/encoding_method.cpp
LinZhihao-723: let's use y# so we don't have to worry about memory manage.

src/clp_ffi/encoding_method.cpp
LinZhihao-723: "To avoid the frequent expansion of ir_buf, allocate sufficient space in advance"

src/clp_ffi/ErrorMessage.hpp
LinZhihao-723: Yes, this one is used for later PR. It shouldn't be added at this point yet.

setup.py
LinZhihao-723: yes, this can be removed.

src/clp_ffi/encoding_method.cpp
LinZhihao-723: If the error happens, it's because the delta range is not handled. The current CLP logging library is returning NotImplemented.

PR #3, opened 2023-07-03
Add basic classes to support IR decoding methods

LinZhihao-723:
# Description
Add two basic classes for IR decoding methods: Metadata and LogEvent
Metadata: a class to represent the decoded preamble
LogEvent: a class to represent a decoded log event (log message + timestamp)
Both classes provide an interface to users in the Python level, and have underlying data stored in C++ layer.

# Validation performed
1. Added unit tests
2. Core functionalities tested and validated with decoding methods implemented in the beta version: https://github.com/LinZhihao-723/clp-ffi-py/tree/ir_decoder
3. Add several clang-tidy checks and all the checks shall pass


davidlion (approved):

setup.py
```diff
@@ -1,33 +1,39 @@
...
-clp_four_byte_encoder: Extension = Extension(
-    name="clp_ffi_py.CLPFourByteEncoder",
+clp_ir: Extension = Extension(
+    name="clp_ffi_py.CLPIR",
     language="c++",
     include_dirs=[
         "src"
     ],
     sources=[
+        "src/clp/components/core/src/TraceableException.cpp",
```
kirkrodrigues: Let's keep these groups alphabetized

requirements-dev.txt
```diff
@@ -2,7 +2,10 @@ black>=22.10.0
 build>=0.8.0
 cibuildwheel>=2.12.3
 clang-format>=16.0.4
+docformatter>=1.6.3
 mypy>=0.982
 mypy-extensions>=0.4.3
 packaging>=21.3
+ruff>=0.0.275
```
kirkrodrigues: Looks like 278 is the latest if you want to use that

requirements-dev.txt
```diff
@@ -2,7 +2,10 @@ black>=22.10.0
 build>=0.8.0
 cibuildwheel>=2.12.3
 clang-format>=16.0.4
+docformatter>=1.6.3
```
kirkrodrigues: Looks like 1.7.5 is the latest if you want to use that

pyproject.toml
```diff
@@ -37,3 +39,13 @@ preview = true
 
 [tool.cibuildwheel.macos]
 archs = ["x86_64", "universal2", "arm64"]
+
+[tool.ruff]
+line-length = 100
+
+[tool.docformatter]
```
kirkrodrigues, line 37: Can we organize the `tool.XX` sections in alphabetical order?

clp_ffi_py/utils.py
```diff
@@ -0,0 +1,33 @@
...
+        timezone = dateutil.tz.UTC
+    dt: datetime = datetime.fromtimestamp(timestamp / 1000, timezone)
+    return dt.isoformat(sep=" ", timespec="milliseconds")
+
+
+def get_timezone_from_timezone_id(timezone_id: str) -> tzinfo:
+    """
+    Gets the Python timezone object of the provided timezone id.
+
+    :param timezone_id: Timezone Id. :raise RuntimeError: If the given
```
kirkrodrigues: Missing a line break before `:raise`?

src/clp_ffi_py/utils.hpp
```diff
@@ -0,0 +1,58 @@
...
+namespace clp_ffi_py {
+/**
+ * Adds the given Python type to the given Python module.
+ * @param new_type Type object to add.
+ * @param type_name Type name identifier.
+ * @param module Python module to hold the new type.
+ * @return true on success.
+ * @return false on failure with the relevant Python exception and error set.
+ */
+auto add_type(PyTypeObject* new_type, char const* type_name, PyObject* module) -> bool;
```
kirkrodrigues: This is a pretty generic method name. How about `add_python_type`?

src/clp_ffi_py/ir/ErrorMessages.hpp
kirkrodrigues, line 1: This filename doesn't follow our naming rules, right? I think it should be `snake_case`.

src/clp_ffi_py/ir/PyLogEvent.hpp
```diff
@@ -0,0 +1,144 @@
...
+    PyMetadata* m_py_metadata;
+};
+
+/**
+ * Gets the PyTypeObject that represents PyLogEvent's Python type. This type is
+ * dynamically created and initialized during the execution of
+ * `PyLogEvent_module_level_init`.
+ * @return Python type object associated with PyLogEvent.
+ */
+auto PyLogEvent_get_PyType() -> PyTypeObject*;
```
kirkrodrigues: Could these methods be static class methods?

src/clp_ffi_py/ir/PyLogEvent.hpp
```diff
@@ -0,0 +1,144 @@
...
+    /**
+     * Validates whether the PyLogEvent has a PyMetadata object associated.
+     * @return Whether `Py_metadata` points to nullptr.
+     */
+    [[nodiscard]] auto has_metadata() -> bool { return nullptr != m_py_metadata; }
+
+    /**
+     * Resets pointers to nullptr.
+     */
+    auto reset() -> void {
```
kirkrodrigues: This docstring and method name are not very descriptive, especially since you have another method called `clean`. Perhaps this should be called `default_init`?

src/clp_ffi_py/ir/PyLogEvent.hpp
```diff
@@ -0,0 +1,144 @@
...
+        m_log_event = nullptr;
+        m_py_metadata = nullptr;
+    }
+
+    /**
+     * Releases the memory allocated for underlying metadata field and the
+     * reference hold for the Python object(s).
+     */
+    auto clean() -> void {
```
kirkrodrigues, line 67: Since this is the counterpart of `init`, can we call it `deinit`?

src/clp_ffi_py/ir/PyLogEvent.hpp
```diff
@@ -0,0 +1,144 @@
+#ifndef CLP_FFI_PY_PY_LOG_EVENT_HPP
```
kirkrodrigues: Typically automated tools will generate the string `CLP_FFI_PY_PYLOGEVENT_HPP`. Can we stick with that convention? cc: @davidlion

src/clp_ffi_py/Py_utils.hpp
```diff
@@ -0,0 +1,37 @@
...
+#include <clp/components/core/src/ffi/encoding_methods.hpp>
+
+namespace clp_ffi_py {
+/**
+ * Initializes CPython interface to Python level utility functions implemented
+ * in submodule `clp_ffi_py.utils`.
+ * @return true on success.
+ * @return false on failure with the relevant Python exception and error set.
+ */
+auto py_utils_init() -> bool;
```
kirkrodrigues, line 15: If you're prefixing every method name, then this feels like you should have a `py_utils` namespace instead. Right, @davidlion?

src/clp_ffi_py/modules/clp_ir.cpp
```diff
@@ -1,11 +1,14 @@
...
-// NOLINTBEGIN(*-avoid-c-arrays)
-// NOLINTBEGIN(cppcoreguidelines-avoid-non-const-global-variables)
-// NOLINTBEGIN(modernize-use-trailing-return-type)
-// NOLINTBEGIN(readability-identifier-naming)
+#include <clp_ffi_py/ir/encoding_methods.hpp>
+#include <clp_ffi_py/ir/PyLogEvent.hpp>
+#include <clp_ffi_py/ir/PyMetadata.hpp>
+#include <clp_ffi_py/Py_utils.hpp>
 
+// CPython macors use C-arrays
```
kirkrodrigues:
```suggestion
// CPython macros use C-arrays
```

src/clp_ffi_py/ir/LogEvent.hpp
```diff
@@ -0,0 +1,79 @@
...
+              m_index{index} {
+        if (formatted_timestamp.has_value()) {
+            m_formatted_timestamp = std::string(formatted_timestamp.value());
+        }
+    }
+
+    [[nodiscard]] auto get_log_message() const -> std::string { return m_log_message; }
+
+    [[nodiscard]] auto get_log_message_view() const -> std::string_view {
+        return std::string_view(m_log_message);
```
kirkrodrigues:
```suggestion
        return std::string_view{m_log_message};
```

src/clp_ffi_py/ir/LogEvent.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+    [[nodiscard]] auto get_formatted_timestamp() const -> std::string {
+        return m_formatted_timestamp;
+    }
+
+    [[nodiscard]] auto get_index() const -> size_t { return m_index; }
+
+    /**
+     * Checks if the log event has the formatted timestamp buffered.
+     * @return Whether the m_formatted_timestamp is empty.
```
kirkrodrigues:
```suggestion
     * @return Whether the log event has the formatted timestamp buffered.
```

* Removes the redundancy between the description and return value.
* Removes the implementation details.

src/clp_ffi_py/ir/Metadata.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include <clp_ffi_py/ir/ErrorMessages.hpp>
+
+namespace clp_ffi_py::ir {
+namespace {
+/**
+ * Validates whether the JSON object contains the given key and has a string
+ * data associated to this particular key.
+ * @param json_data JSON object to be validated.
+ * @param key The key to access the data field.
+ * @return Return true if the data is valid.
```
kirkrodrigues:
```suggestion
 * @return true if the data is valid.
```

src/clp_ffi_py/ir/Metadata.cpp
```diff
@@ -0,0 +1,75 @@
...
+        );
+    }
+    try {
+        std::string const ref_timestamp_str{metadata[ref_timestamp_key]};
+        m_ref_timestamp = static_cast<ffi::epoch_time_ms_t>(std::stoull(ref_timestamp_str));
+    } catch (std::exception const& ex) {
+        throw ExceptionFFI(ErrorCode_Unsupported, __FILE__, __LINE__, ex.what());
+    }
+
+    auto const* const timestamp_format_key{
```
kirkrodrigues: The second `const` seems excessive, but nbd if you want to keep it.

src/clp_ffi_py/ir/Metadata.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Constructs a new Metadata object from the provided fields. Currently,
+     * `m_is_four_byte_encoding` is set to true by default since it is the only
+     * format supported.
+     * @param ref_timestamp The reference timestamp used to calculate the
+     * timestamp of the first log message in the IR stream.
+     * @param timestamp_format Timestamp format to be use when generating the
+     * logs with a reader.
+     * @param timezone Timezone in TZID format to be use when generating the
```
kirkrodrigues:
```suggestion
     * @param timestamp_format Timestamp format to use when generating the logs
     * with a reader.
     * @param timezone Timezone in TZID format to use when generating the
```

tests/test_clpir.py
```diff
@@ -0,0 +1,280 @@
...
+from datetime import tzinfo
+from typing import Optional
+
+
+class TestCaseMetadata(unittest.TestCase):
+    """
+    Class for testing clp_ffi_py.CLPIRDecoder.Metadata.
+    """
+
+    def check_metadata(
```
kirkrodrigues: Ideally make these helper methods private (prefix with `__`) and move them under the tests according to our ordering rules (public before private, etc.).

kirkrodrigues (changes requested):

src/clp_ffi_py/Py_utils.hpp
LinZhihao-723: The only reason we have `py_` here is to follow the convention of C Python, where the python level API should start with `py_`

src/clp_ffi_py/ir/PyLogEvent.hpp
LinZhihao-723: I would prefer `clean`. `deinit` sounds weird and I don't think it makes a large difference here.

src/clp_ffi_py/ir/PyLogEvent.hpp
LinZhihao-723: ideally, it could. But conceptually the actual type object returned by this function is not `PyLogEvent` object. It's a type object dynamically created during the module initialization, and not directly correlated to the `PyLogEvent` object. I would prefer to keep them separate.

kirkrodrigues (approved):

davidlion (approved):

PR #4, opened 2023-07-19
Implements the basis for Python search query used by the decoding methods

LinZhihao-723:
# Description
Implements the basis of the search query, which takes a timestamp range and a list of wildcard queries as the user input. This class can determine whether a log event or message matches the search requirement, which will be applied to the IR stream decoder to filter out irrelevant log events.

# Validation performed
1. This class has yet to be used by any code. It does not affect the correctness of the unit tests.
2. The implementation was tested with the functional decoding methods in the beta version.


src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,198 @@
...
+     * @param wildcard Wildcard string.
+     * @param case_sensitive Case sensitive indicator.
+     */
+    WildcardQuery(std::string wildcard, bool case_sensitive)
+            : m_wildcard(std::move(wildcard)),
+              m_case_sensitive(case_sensitive){};
+
+    [[nodiscard]] auto get_wildcard() const -> std::string_view { return m_wildcard; }
+
+    [[nodiscard]] auto get_case_sensitive() const -> bool { return m_case_sensitive; }
```
davidlion:
```suggestion
    [[nodiscard]] auto is_case_sensitive() const -> bool { return m_case_sensitive; }
```

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,198 @@
...
+private:
+    std::string m_wildcard;
+    bool m_case_sensitive;
+};
+
+/**
+ * This class represents a search query, utilized for filtering log events in a
+ * CLP IR stream. The query could include a list of wildcard searches aimed at
+ * identifying certain log messages, and a timestamp interval with a lower and
+ * upper bound.  This class provides an interface to set up a search query, as
```
davidlion:
```suggestion
 * upper bound. This class provides an interface to set up a search query, as
```

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,198 @@
...
+    /**
+     * Ideally, when searching an IR stream with a query, the search terminates
+     * once the timestamp exceeds the search upper bound. However, the timestamp
+     * might not be monotonically increasing in a CLP IR stream. It can be
+     * locally disordered due to thread contention. To safely stop searching, we
+     * need to ensure that the current timestamp has exceeded the search's upper
+     * bound by a reasonable margin. This margin defaults to the following
+     * constant. During the query initialization, it will be applied to the
+     * upper bound timestamp to generate a timestamp which indicates the safe
+     * termination.
```
davidlion:
```suggestion
     * upper bound timestamp to generate a timestamp which indicates it is safe
     * to terminate the search.
```

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,198 @@
...
+     */
+    static constexpr ffi::epoch_time_ms_t const cDefaultSearchTerminationMargin{
+            static_cast<ffi::epoch_time_ms_t>(60 * 1000)};
+
+    /**
+     * Constructs an empty query object: the wildcard list is empty and the
+     * timestamp range is set to include all the valid Unix epoch time.
```
davidlion:
```suggestion
     * Constructs an empty query object that will match all logs. The wildcard
     * list is empty and the timestamp range is set to include all the valid
     * Unix epoch timestamps.
```

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    [[nodiscard]] auto get_ts_upper_bound_ts() const -> ffi::epoch_time_ms_t {
+        return m_upper_bound_ts;
+    }
+
+    /**
+     * @param ts Input timestamp.
+     * @return true if the given timestamp is in the search time rage bounded by
+     * the lower bound and the upper bound timestamp (inclusive).
```
davidlion:
```suggestion
     * @return true if the given timestamp is in the search time range bounded
     * by the lower bound and the upper bound timestamp (inclusive).
```

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,198 @@
...
+    bool m_case_sensitive;
+};
+
+/**
+ * This class represents a search query, utilized for filtering log events in a
+ * CLP IR stream. The query could include a list of wildcard searches aimed at
+ * identifying certain log messages, and a timestamp interval with a lower and
+ * upper bound.  This class provides an interface to set up a search query, as
+ * well as methods to validate whether the query can be matched by a log event.
+ */
```
davidlion:
```suggestion
 * well as methods to validate whether the query can be matched by a log event.
 * Note that an empty wildcard query list will match any log within the range.
 */
```

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,198 @@
...
+
+private:
+    std::string m_wildcard;
+    bool m_case_sensitive;
+};
+
+/**
+ * This class represents a search query, utilized for filtering log events in a
+ * CLP IR stream. The query could include a list of wildcard searches aimed at
+ * identifying certain log messages, and a timestamp interval with a lower and
```
davidlion:
```suggestion
 * identifying certain log messages, and a timestamp range with a lower and
```

davidlion (changes requested):

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,200 @@
...
+class WildcardQuery {
+public:
+    /**
+     * Initializes the wildcard query.
+     * @param wildcard Wildcard string.
+     * @param case_sensitive Case sensitive indicator.
+     */
+    WildcardQuery(std::string wildcard, bool case_sensitive)
+            : m_wildcard(std::move(wildcard)),
+              m_case_sensitive(case_sensitive){};
```
kirkrodrigues, line 29:
```suggestion
              m_case_sensitive(case_sensitive) {};
```

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,200 @@
...
+
+    [[nodiscard]] auto get_wildcard() const -> std::string_view { return m_wildcard; }
+
+    [[nodiscard]] auto is_case_sensitive() const -> bool { return m_case_sensitive; }
+
+private:
+    std::string m_wildcard;
```
kirkrodrigues: How about `m_wildcard_query`? A wildcard is technically `*` or `?`.

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,200 @@
...
+namespace clp_ffi_py::ir {
+/**
+ * This class defines a wildcard query, which includes a wildcard string and a
+ * boolean value to indicate if the match is case-sensitive.
+ */
+class WildcardQuery {
+public:
+    /**
+     * Initializes the wildcard query.
+     * @param wildcard Wildcard string.
```
kirkrodrigues: Let's add a description about what wildcards are supported. In our experience, not everyone is clear about what a wildcard is.

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,200 @@
...
+     */
+    static constexpr ffi::epoch_time_ms_t const cDefaultSearchTerminationMargin{
+            static_cast<ffi::epoch_time_ms_t>(60 * 1000)};
+
+    /**
+     * Constructs an empty query object that will match all logs. The wildcard
+     * list is empty and the timestamp range is set to include all the valid
+     * Unix epoch timestamps.
+     */
+    explicit Query()
```
kirkrodrigues, line 70: Is `explicit` necessary?

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,200 @@
...
+                              : cTimestampMax},
+              m_wildcard_list{std::move(wildcard_list)} {
+        throw_if_ts_range_invalid();
+    }
+
+    [[nodiscard]] auto get_lower_bound_ts() const -> ffi::epoch_time_ms_t {
+        return m_lower_bound_ts;
+    }
+
+    [[nodiscard]] auto get_ts_upper_bound_ts() const -> ffi::epoch_time_ms_t {
```
kirkrodrigues:
```suggestion
    [[nodiscard]] auto get_upper_bound_ts() const -> ffi::epoch_time_ms_t {
```

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,200 @@
...
+     */
+    [[nodiscard]] auto terminate_search(ffi::epoch_time_ms_t ts) const -> bool {
+        return m_search_termination_ts <= ts;
+    }
+
+    /**
+     * Validates whether the input log message matches any of the wildcard
+     * conditions for the query.
+     * @param log_message Input log message.
+     * @return true if the wildcard list is empty or has at least one match.
```
kirkrodrigues:
```suggestion
     * @return true if the wildcard query list is empty or at least one wildcard query matches.
```

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,200 @@
...
+                    __LINE__,
+                    "Search query lower bound timestamp exceeds the upper bound timestamp."
+            );
+        }
+    }
+
+    ffi::epoch_time_ms_t m_lower_bound_ts;
+    ffi::epoch_time_ms_t m_upper_bound_ts;
+    ffi::epoch_time_ms_t m_search_termination_ts;
+    std::vector<WildcardQuery> m_wildcard_list;
```
kirkrodrigues:
```suggestion
    std::vector<WildcardQuery> m_wildcard_queries;
```

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,200 @@
...
+     * termination timestamp.
+     * @return false otherwise.
+     */
+    [[nodiscard]] auto terminate_search(ffi::epoch_time_ms_t ts) const -> bool {
+        return m_search_termination_ts <= ts;
+    }
+
+    /**
+     * Validates whether the input log message matches any of the wildcard
+     * conditions for the query.
```
kirkrodrigues:
```suggestion
     * Validates whether the input log message matches any of the wildcard
     * queries in the query.
```

davidlion (approved):

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,200 @@
...
+
+    /**
+     * Determines whether the search can terminate by evaluating the input
+     * timestamp.
+     * @param ts Input timestamp.
+     * @return true if the given timestamp is equal to or greater than the
+     * termination timestamp.
+     * @return false otherwise.
+     */
+    [[nodiscard]] auto terminate_search(ffi::epoch_time_ms_t ts) const -> bool {
```
kirkrodrigues:
```suggestion
    /**
     * @param ts Input timestamp.
     * @return Whether the given timestamp is safely outside this query's time
     * range.
     */
    [[nodiscard]] auto ts_safely_outside_time_range(ffi::epoch_time_ms_t ts) const -> bool {
```

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,200 @@
...
+     * Ideally, when searching an IR stream with a query, the search terminates
+     * once the timestamp exceeds the search upper bound. However, the timestamp
+     * might not be monotonically increasing in a CLP IR stream. It can be
+     * locally disordered due to thread contention. To safely stop searching, we
+     * need to ensure that the current timestamp has exceeded the search's upper
+     * bound by a reasonable margin. This margin defaults to the following
+     * constant. During the query initialization, it will be applied to the
+     * upper bound timestamp to generate a timestamp which indicates it is safe
+     * to terminate the search.
+     */
```
kirkrodrigues:
```suggestion
    /**
     * Ideally, when searching an IR stream, the search would terminate once the
     * current log event's timestamp exceeds the query's time range. However,
     * timestamps in the IR stream might not be monotonically increasing; they
     * can be locally disordered due to thread contention. So to safely stop the
     * search, we need to ensure the current timestamp in the IR stream exceeds
     * the query's time range by a reasonable margin. This margin defaults to
     * the following constant. During the query initialization, it will be
     * applied to the upper bound timestamp to generate a timestamp which
     * indicates it is safe to terminate the search.
     */
```

kirkrodrigues: I think the description of the problem should go in the class' docstring, since it's relevant in multiple methods, not just in the specification of this constant.

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,200 @@
...
+     * empty wildcard list.
+     * @param search_time_lower_bound Start of search time range (inclusive).
+     * @param search_time_upper_bound End of search time range (inclusive).
+     * @param search_termination_margin The margin used to determine the search
+     * termination timestamp.
+     */
+    explicit Query(
+            ffi::epoch_time_ms_t search_time_lower_bound,
+            ffi::epoch_time_ms_t search_time_upper_bound,
+            ffi::epoch_time_ms_t search_termination_margin = cDefaultSearchTerminationMargin
```
kirkrodrigues: How about `search_time_termination_margin`?

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,200 @@
...
+    /**
+     * Constructs a new query object with the given timestamp range and a
+     * wildcard list.
+     * @param search_time_lower_bound Start of search time range (inclusive).
+     * @param search_time_upper_bound End of search time range (inclusive).
+     * @param wildcard_list A reference to the wildcard queries vector, whose
+     * data will be transferred using std::move to initialize m_wildcard_list.
+     * @param search_termination_margin The margin used to determine the search
+     * termination timestamp. By default, this value is set to
+     * cDefaultSearchTerminationMargin.
```
kirkrodrigues: Don't think you need to specify the default value since any decent doc site generator will pick up the default value from the code.

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,200 @@
...
+    [[nodiscard]] auto get_ts_upper_bound_ts() const -> ffi::epoch_time_ms_t {
+        return m_upper_bound_ts;
+    }
+
+    /**
+     * @param ts Input timestamp.
+     * @return true if the given timestamp is in the search time range bounded
+     * by the lower bound and the upper bound timestamp (inclusive).
+     */
+    [[nodiscard]] auto ts_in_range(ffi::epoch_time_ms_t ts) const -> bool {
```
kirkrodrigues: For the three match methods, how about:
* `matches_time_range`
* `matches_wildcard_queries`
* `matches`

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,200 @@
...
+     * @param log_event Input log event.
+     * @return true if the timestamp is in range, and the wildcard list is empty
+     * or has at least one match.
+     * @return false otherwise.
+     */
+    [[nodiscard]] auto matches(LogEvent const& log_event) const -> bool {
+        if (false == ts_in_range(log_event.get_timestamp())) {
+            return false;
+        }
+        return wildcard_matches(log_event.get_log_message_view());
```
kirkrodrigues:
```suggestion
        return ts_in_range(log_event.get_timestamp())
               && wildcard_matches(log_event.get_log_message_view());
```

src/clp_ffi_py/ir/Query.cpp
```diff
@@ -0,0 +1,25 @@
...
+namespace clp_ffi_py::ir {
+auto Query::wildcard_matches(std::string_view log_message) const -> bool {
+    if (m_wildcard_list.empty()) {
+        return true;
+    }
+    return std::any_of(
+            m_wildcard_list.begin(),
+            m_wildcard_list.end(),
+            [&](auto const& wildcard_query) {
+                return wildcard_match_unsafe(
```
kirkrodrigues, line 17: Where will the validation for the input wildcard queries go (in order to use this method)?

src/clp_ffi_py/ir/Query.hpp
kirkrodrigues: We should also change all instances of wildcard to wildcard query (unless you actually mean wildcard).

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,200 @@
...
+     * Constructs a new query object with the given timestamp range with an
+     * empty wildcard list.
+     * @param search_time_lower_bound Start of search time range (inclusive).
+     * @param search_time_upper_bound End of search time range (inclusive).
+     * @param search_termination_margin The margin used to determine the search
+     * termination timestamp.
+     */
+    explicit Query(
+            ffi::epoch_time_ms_t search_time_lower_bound,
+            ffi::epoch_time_ms_t search_time_upper_bound,
```
kirkrodrigues, line 85: How about `time_range_begin` and `time_range_end`?

kirkrodrigues (changes requested):

src/clp_ffi_py/ir/Query.hpp
LinZhihao-723: This one is omitted by clang-format

src/clp_ffi_py/ir/Query.hpp
LinZhihao-723: Since this file is not user-oriented, can we add the description in PyQuery instead?

kirkrodrigues: Sure, I'm fine with putting the clear descriptions in the corresponding Python layer.

src/clp_ffi_py/ir/Query.cpp
LinZhihao-723: when parsing the user input in py `__init__`

src/clp_ffi_py/ir/Query.hpp
LinZhihao-723: I would stick to `lower/upper bound`

src/clp_ffi_py/ir/Query.cpp
kirkrodrigues: Gotcha, then we should add a line in the constructors to say something like `the wildcard query must be valid (see `wildcard_match_unsafe`).

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,192 @@
...
+ */
+class WildcardQuery {
+public:
+    /**
+     * Initializes the wildcard query.
+     * @param wildcard Wildcard string.
+     * @param case_sensitive Case sensitive indicator.
+     */
+    WildcardQuery(std::string wildcard, bool case_sensitive)
+            : m_wildcard_query(std::move(wildcard)),
```
kirkrodrigues:
```suggestion
     * @param wildcard_query Wildcard query.
     * @param case_sensitive Case sensitive indicator.
     */
    WildcardQuery(std::string wildcard_query, bool case_sensitive)
            : m_wildcard_query(std::move(wildcard_query)),
```

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,192 @@
...
+ * Ideally, when searching an IR stream with a query, the search terminates
+ * once the timestamp exceeds the search upper bound. However, the timestamp
+ * might not be monotonically increasing in a CLP IR stream. It can be
+ * locally disordered due to thread contention. To safely stop searching, we
+ * need to ensure that the current timestamp has exceeded the search's upper
+ * bound by a reasonable margin. This margin defaults to the following
+ * constant. During the query initialization, it will be applied to the
+ * upper bound timestamp to generate a timestamp which indicates it is safe
+ * to terminate the search.
+ */
```
kirkrodrigues:
```suggestion
/**
 * This class represents a search query, utilized for filtering log events in a
 * CLP IR stream. The query could include a list of wildcard queries aimed at
 * identifying certain log messages, and a timestamp range with a lower and
 * upper bound. This class provides an interface to set up a search query, as
 * well as methods to validate whether the query can be matched by a log event.
 * Note that an empty wildcard query list will match any log within the range.
 * <p>
 * NOTE: When searching an IR stream with a query, ideally, the search
 * would terminate once the current log event's timestamp exceeds the upper
 * bound of the query's time range. However, the timestamps in the IR stream
 * might not be monotonically increasing; they can be locally disordered due to
 * thread contention. So to safely stop searching, we need to ensure that the
 * current timestamp in the IR stream exceeds the query's upper bound timestamp
 * by a reasonable margin. This margin can be specified by the user or it
 * will default to `cDefaultSearchTerminationMargin`.
 */
```

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,192 @@
...
+    static constexpr ffi::epoch_time_ms_t const cTimestampMin{0};
+    static constexpr ffi::epoch_time_ms_t const cTimestampMax{
+            std::numeric_limits<ffi::epoch_time_ms_t>::max()};
+    static constexpr ffi::epoch_time_ms_t const cDefaultSearchTerminationMargin{
+            static_cast<ffi::epoch_time_ms_t>(60 * 1000)};
+
+    /**
+     * Constructs an empty query object that will match all logs. The wildcard
+     * list is empty and the timestamp range is set to include all the valid
+     * Unix epoch timestamps.
```
kirkrodrigues:
```suggestion
     * query list is empty and the timestamp range is set to include all the
     * valid Unix epoch timestamps.
```

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,192 @@
...
+              m_upper_bound_ts{cTimestampMax},
+              m_search_termination_ts{cTimestampMax} {};
+
+    /**
+     * Constructs a new query object with the given timestamp range with an
+     * empty wildcard list.
+     * @param search_time_lower_bound Start of search time range (inclusive).
+     * @param search_time_upper_bound End of search time range (inclusive).
+     * @param search_time_termination_margin The margin used to determine the
+     * search termination timestamp.
```
kirkrodrigues:
```suggestion
     * search termination timestamp (see note in the class' docstring).
```

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,192 @@
...
+              m_search_termination_ts{
+                      (cTimestampMax - search_time_termination_margin > search_time_upper_bound)
+                              ? search_time_upper_bound + search_time_termination_margin
+                              : cTimestampMax} {
+        throw_if_ts_range_invalid();
+    }
+
+    /**
+     * Constructs a new query object with the given timestamp range and a
+     * wildcard list.
```
kirkrodrigues:
```suggestion
     * wildcard query list.
```

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,192 @@
...
+     * @param search_time_lower_bound Start of search time range (inclusive).
+     * @param search_time_upper_bound End of search time range (inclusive).
+     * @param wildcard_list A reference to the wildcard queries vector, whose
+     * data will be transferred using std::move to initialize
```
kirkrodrigues:
```suggestion
     * @param wildcard_queries A list of wildcard queries. Each wildcard query must
     * be valid (see `wildcard_match_unsafe`).
```

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,192 @@
...
+     * m_wildcard_queries. The given wildcard query must be valid (see
+     * wildcard_match_unsafe).
+     * @param search_time_termination_margin The margin used to determine the
+     * search termination timestamp.
```
kirkrodrigues:
```suggestion
     * search termination timestamp (see note in the class' docstring).
```

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,192 @@
...
+          ffi::epoch_time_ms_t search_time_upper_bound,
+          std::vector<WildcardQuery> wildcard_list,
+          ffi::epoch_time_ms_t search_time_termination_margin = cDefaultSearchTerminationMargin)
+            : m_lower_bound_ts{search_time_lower_bound},
+              m_upper_bound_ts{search_time_upper_bound},
+              m_search_termination_ts{
+                      (cTimestampMax - search_time_termination_margin > search_time_upper_bound)
+                              ? search_time_upper_bound + search_time_termination_margin
+                              : cTimestampMax},
+              m_wildcard_queries{std::move(wildcard_list)} {
```
kirkrodrigues:
```suggestion
          std::vector<WildcardQuery> wildcard_queries,
          ffi::epoch_time_ms_t search_time_termination_margin = cDefaultSearchTerminationMargin)
            : m_lower_bound_ts{search_time_lower_bound},
              m_upper_bound_ts{search_time_upper_bound},
              m_search_termination_ts{
                      (cTimestampMax - search_time_termination_margin > search_time_upper_bound)
                              ? search_time_upper_bound + search_time_termination_margin
                              : cTimestampMax},
              m_wildcard_queries{std::move(wildcard_queries)} {
```

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,192 @@
...
+     * by the lower bound and the upper bound timestamp (inclusive).
+     */
+    [[nodiscard]] auto matches_time_range(ffi::epoch_time_ms_t ts) const -> bool {
+        return m_lower_bound_ts <= ts && ts <= m_upper_bound_ts;
+    }
+
+    /**
+     * @param ts Input timestamp.
+     * @return Whether the given timestamp is safely outside this query's time
+     * range.
```
kirkrodrigues:
```suggestion
     * range (see note in the class' docstring).
```

src/clp_ffi_py/ir/Query.hpp
```diff
@@ -0,0 +1,192 @@
...
+    [[nodiscard]] auto ts_safely_outside_time_range(ffi::epoch_time_ms_t ts) const -> bool {
+        return m_search_termination_ts <= ts;
+    }
+
+    /**
+     * Validates whether the input log message matches any of the wildcard
+     * queries in the query.
+     * @param log_message Input log message.
+     * @return true if the wildcard query list is empty or has at least one
+     * wildcard query matches.
```
kirkrodrigues:
```suggestion
     * @return true if the wildcard query list is empty or at least one wildcard
     * query matches.
```

kirkrodrigues (changes requested):

kirkrodrigues (approved):

PR #7, opened 2023-07-24
Add version numbers in setup.py

LinZhihao-723:
# Description
The version number must be properly added to the `setup.py` so that Python3.6 wheel can be built properly by `cibuildwheel`. Ideally, with this change added we will no longer need to build for Python3.6 manually.
In addition, this PR also automates the build environment setup when building for macOS. By default, most of macOS environment has `MACOSX_DEPLOYMENT_TARGET ` set to `10.13`, which does not support c++17 compilation. With this PR, the build tool will automatically set `MACOSX_DEPLOYMENT_TARGET ` to `10.15` (the minimum requirement to build c++17 features) if the default version doesn't support c++17.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,4 +37,5 @@ setup(
     description="CLP FFI Python Interface",
     ext_modules=[ir],
     packages=["clp_ffi_py"],
+    version="0.0.1"
 )
--- a/setup.py
+++ b/setup.py
@@ -1,19 +1,19 @@
 import os
+import sys
+import toml
 from setuptools import setup, Extension
+from typing import Any, Dict, Optional
 
 ir: Extension = Extension(
     name="clp_ffi_py.ir",
     language="c++",
-    include_dirs=[
-        "src"
-    ],
+    include_dirs=["src"],
     sources=[
         "src/clp/components/core/src/ffi/ir_stream/decoding_methods.cpp",
         "src/clp/components/core/src/ffi/ir_stream/encoding_methods.cpp",
         "src/clp/components/core/src/ffi/encoding_methods.cpp",
         "src/clp/components/core/src/string_utils.cpp",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -37,4 +37,5 @@
     description="CLP FFI Python Interface",
     ext_modules=[ir],
     packages=["clp_ffi_py"],
+    version="0.0.1"
```
kirkrodrigues: Probably worth adding a comment in pyproject.toml that the packages and version specification also need to be specified in `setup.py` (and why)

setup.py
```diff
@@ -24,17 +24,31 @@
...
 
-setup(
-    name="clp_ffi_py",
-    description="CLP FFI Python Interface",
-    ext_modules=[ir],
-    packages=["clp_ffi_py"],
-)
+if "__main__" == __name__:
+    try:
+        if hasattr(os, "sys") and "darwin" == os.sys.platform:
```
kirkrodrigues: Why not use `platform.system()`?

setup.py
```diff
@@ -24,17 +24,31 @@
...
+            target: Optional[str] = os.environ.get("MACOSX_DEPLOYMENT_TARGET")
+            if None is target or float(target) < 10.15:
+                os.environ["MACOSX_DEPLOYMENT_TARGET"] = "10.15"
+
+        config: Dict[str, Any] = toml.load("pyproject.toml")
+        version: Optional[str] = config.get("project", {}).get("version", None)
+        if None is version:
+            sys.exit("Error: The version number is not found in pyproject.toml")
```
kirkrodrigues:
```suggestion
            sys.exit("Error: The version number was not found in pyproject.toml")
```

kirkrodrigues (approved):

PR #8, opened 2023-07-28
Implements PyQuery

LinZhihao-723:
# Description
Implements Query (using C extension) and WildcardQuery (in Python) as Python objects that provides a Python interface for users to specify an IR search query.

# Validation performed
1. Added a set of new unit tests to verify the functionality of the Query object, including initialization, serialization, and log event matching.
2. This object is implemented and tested in the beta version FFI library.


src/clp_ffi_py/utils.hpp
```diff
@@ -40,6 +40,13 @@ auto parse_py_string(PyObject* py_string, std::string& out) -> bool;
  */
 auto parse_py_string_as_string_view(PyObject* py_string, std::string_view& view) -> bool;
 
+/**
+ * Gets the Python True/False from a given `bool` value/expression.
```
davidlion:
```suggestion
 * Gets the Python True/False object from a given `bool` value/expression.
```

src/clp_ffi_py/ir/PyMetadata.hpp
```diff
@@ -11,9 +11,9 @@
 namespace clp_ffi_py::ir {
 /**
  * A PyObject structure functioning as a Python-compatible interface to retrieve
- * CLP IR metadata. The underlying data is pointed by `metadata`. Additionally,
- * it retains a tzinfo object at the Python level that signifies the
- * corresponding timezone.
+ * CLP IR metadata. The underlying data is pointed by `m_metadata`.
```
davidlion:
```suggestion
 * CLP IR metadata. The underlying data is pointed to by `m_metadata`.
```

src/clp_ffi_py/ir/PyLogEvent.hpp
```diff
@@ -12,9 +12,9 @@
 namespace clp_ffi_py::ir {
 /**
  * A PyObject structure functioning as a Python-compatible interface to retrieve
- * a log event. The underlying data is pointed by `log_event`. The object may
- * reference a PyMetadata object pointed to by `Py_metadata` that specifies the
- * event's metadata, such as timestamp format, from the preamble.
+ * a log event. The underlying data is pointed by `m_log_event`. The object may
+ * reference a PyMetadata object pointed to by `m_py_metadata` that specifies
+ * the event's metadata, such as timestamp format, from the preamble.
```
davidlion:
```suggestion
 * a log event. The underlying data is pointed to by `m_log_event`. The object
 * may reference a PyMetadata object pointed to by `m_py_metadata` that
 * specifies the event's metadata, such as timestamp format, from the preamble.
```

clp_ffi_py/ir.pyi
```diff
@@ -25,6 +26,29 @@ class LogEvent:
...
+        search_time_upper_bound: int = default_search_time_upper_bound(),
+        wildcard_queries: Optional[List[WildcardQuery]] = None,
+        search_time_termination_margin: int = default_search_time_termination_margin(),
+    ): ...
+    def __getstate__(self) -> Dict[str, Any]: ...
+    def __setstate__(self, state: Dict[str, Any]) -> None: ...
+    def get_search_time_lower_bound(self) -> int: ...
+    def get_search_time_upper_bound(self) -> int: ...
+    def get_wildcard_queries(self) -> Optional[List[WildcardQuery]]: ...
+    def get_search_time_termination_margin(self) -> int: ...
```
davidlion:
```suggestion
    def get_search_time_termination_margin(self) -> int: ...
    def get_wildcard_queries(self) -> Optional[List[WildcardQuery]]: ...
```

src/clp_ffi_py/ir/PyLogEvent.cpp
```diff
@@ -291,6 +292,27 @@ auto PyLogEvent_get_index(PyLogEvent* self) -> PyObject* {
     return PyLong_FromLongLong(static_cast<int64_t>(self->get_log_event()->get_index()));
 }
 
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays)
+PyDoc_STRVAR(
+        cPyLogEventMatchQueryDoc,
+        "match_query(self, query)\n"
+        "--\n\n"
+        "Matches the underlying log event against the given query. Check the document of the Query "
+        "object for more details.\n"
```
davidlion:
```suggestion
        "Matches the underlying log event against the given query. Refer to the documentation of "
        "clp_ffi_py.Query for more details.\n"
```

clp_ffi_py/query.py
davidlion, line 1: This file should be named wildcard_query.py, since the actual query class is really ir.query (which I think/agree makes sense).

src/clp_ffi_py/ir/PyQuery.hpp
```diff
@@ -0,0 +1,87 @@
...
+#include <clp_ffi_py/PyObjectUtils.hpp>
+
+namespace clp_ffi_py::ir {
+class PyQuery {
+    /**
+     * A PyObject structure functioning as a Python-compatible interface to retrieve
+     * a search query that is used to filtering log events in a CLP IR stream. The
+     * underlying data is pointed by `m_query`. The detailed description can be
+     * found in the PyQuery Python doc string.
+     */
```
davidlion:
```suggestion
/**
 * A PyObject structure functioning as a Python-compatible interface to retrieve
 * a search query that is used to filter log events in a CLP IR stream. The
 * underlying data is pointed to by `m_query`. A detailed description can be
 * found in the PyQuery Python doc strings.
 */
class PyQuery {
```

src/clp_ffi_py/ir/PyQuery.hpp
```diff
@@ -0,0 +1,87 @@
...
+public:
+    /**
+     * Initializes the underlying data with the given input.
+     * Since the memory allocation of PyQuery is handled by CPython's
+     * allocator, cpp constructor will ot be explicitly called. This function
+     * serves as the default constructor to initialize the underlying query.
+     * It has to be manually called whenever creating a new PyQuery object
+     * through CPython AIPs.
```
davidlion:
```suggestion
     * allocator, any cpp constructor will not be explicitly called. This
     * function serves as the default constructor to initialize the underlying
     * query. It has to be manually called whenever creating a new PyQuery
     * object through CPython AIPs.
```

tests/test_clpir.py
```diff
@@ -315,5 +315,434 @@ def test_encoding_methods_consistency(self) -> None:
         self.assertEqual(encoded_message_and_ts_delta, encoded_message + encoded_ts_delta)
 
 
+class TestCaseQueryBase(unittest.TestCase):
+    """
+    Base class for Query related testing.
+
+    Helper functions should be defined in this class. This base class shouldn't
+    be launched during the testing.
```
```suggestion
@unittest.skip("Base class should not be used during testing.")
class TestCaseQueryBase(unittest.TestCase):
    """
    Base class for Query related testing.

    Helper functions should be defined in this class. 
```

Although currently there are no actual tests to run in this base class, it is better to explicitly guard against this.

tests/test_clpir.py
```diff
@@ -315,5 +315,434 @@ def test_encoding_methods_consistency(self) -> None:
...
+        range.
+        """
+        query: Query
+        search_time_lower_bound: int
+        search_time_upper_bound: int
+        search_time_termination_margin: int
+        exception_captured: bool
+
+        # Note: for the default initialization, the actual search time
+        # termination margin should be set to 0, otherwise it will overflow
```
```suggestion
        # termination margin should have be set to 0, otherwise it will overflow
```

src/clp_ffi_py/ir/PyQuery.cpp
```diff
@@ -0,0 +1,650 @@
...
+    static char* keyword_table[]{
+            static_cast<char*>(keyword_search_time_lower_bound),
+            static_cast<char*>(keyword_search_time_upper_bound),
+            static_cast<char*>(keyword_wildcard_queries),
+            static_cast<char*>(keyword_search_time_termination_margin),
+            nullptr};
+
+    // If the argument parsing fails, `self` will be deallocated. We must reset
+    // all pointers to nullptr in advance, otherwise the deallocator might
+    // trigger segmentation fault
```
davidlion:
```suggestion
    // trigger a segmentation fault.
```

src/clp_ffi_py/ir/PyQuery.cpp
```diff
@@ -0,0 +1,650 @@
...
+PyDoc_STRVAR(
+        cPyQueryDoc,
+        "This class represents a search query, utilized for filtering log events in a CLP IR "
+        "stream. The query could include a list of wildcard queries aimed at identifying certain "
+        "log messages, and a timestamp range with a lower and upper bound. This class provides an "
+        "interface to set up a search query, as well as methods to validate whether the query can "
+        "be matched by a log event. Note that an empty wildcard query list will match any log "
+        "within the range.\n"
+        "By default, the wildcard query list is empty and the timestamp range is set to include "
+        "all the valid Unix epoch timestamps. To filtering certain log messages, use customized "
```
davidlion:
```suggestion
        "all the valid Unix epoch timestamps. To filter certain log messages, use customized "
```

davidlion (changes requested):
See the individual comments that have been made prior to this review.

davidlion (approved):

README.md
```diff
@@ -47,6 +47,11 @@ pip install -e .
 python -m unittest -bv
 ```
 
+Note: the testing procedure showing above only works when installing the package using
+`pip install -e .`. If the package is installed from a whl file into the site package, the tester cannot be launched from
+the project's root directory because the `clp_ffi_py` module import conflicts with the directory
+`clp_ffi_py`.
```
davidlion:
```suggestion
Note: If the package is installed from a `whl` file into the site packages,
rather than installed locally (`pip install -e .`),
the tester cannot be launched from the project's root directory.
If `unittest` is ran from the root directory,
the local `clp_ffi_py` directory will shadow the `clp_ffi_py` module installed.
```

PR #9, opened 2023-07-31
Implements PyDecoderBuffer

LinZhihao-723:
# Description
Implements Python decoder buffer. It behaves similarly to what's already implemented in CLP Python logging lib's decoding methods. By specifying a Python buffer protocol, this object can be used to read from a Python IO[bytes] stream directly.

# Validation performed
1. The similar code has been justified in the beta version.
2. Implemented unit tests to simulate what the decoding methods would operate on this buffer.
3. All the linting checks passed.


src/clp_ffi_py/ir/PyDecoderBuffer.cpp
```diff
@@ -0,0 +1,316 @@
...
+                return nullptr;
+            }
+            if (0 == num_bytes_read_from_istream) {
+                reach_istream_end = true;
+            }
+            num_bytes_to_read = std::min<Py_ssize_t>(num_bytes_to_read, m_buffer_size);
+        }
+        auto* unconsumed_bytes{get_unconsumed_bytes()};
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+        read_bytes.insert(read_bytes.end(), unconsumed_bytes, unconsumed_bytes + num_bytes_to_read);
```
davidlion: This appears to only rely on new code and isn't directly limited by cpython. Why are we disabling the linting for it?

tests/test_clpir.py
```diff
@@ -744,5 +762,101 @@ def test_log_event_match(self) -> None:
...
+            decoder buffer.
+        """
+        current_dir: Path = Path(__file__).resolve().parent
+        test_src_dir: Path = current_dir / TestCaseDecoderBuffer.input_src_dir
+        for file_path in test_src_dir.rglob("*"):
+            if not file_path.is_file():
+                continue
+            streaming_result: bytearray
+            decoder_buffer: DecoderBuffer
+            random_seed: int = floor(time.time() * 1000) % 1000
```
davidlion: You might as well generate a "proper" seed (https://docs.python.org/3/library/random.html)
and do it each iteration rather than just adding 1.

LinZhihao-723: Thanks. It's good to know that python `random` methods will use the system time as the seed by default (I thought it is the same as C/C++ that defaults to 0)

src/clp_ffi_py/ir/PyDecoderBuffer.hpp
```diff
@@ -0,0 +1,183 @@
+#ifndef CLP_FFI_PY_PYDECODERBUFFER_HPP
+#define CLP_FFI_PY_PYDECODERBUFFER_HPP
+
+#include <clp_ffi_py/Python.hpp>  // Must always be included before any other header files
+
+#include <GSL/include/gsl/span>
```
davidlion: Need to add `|gsl` to 3rd parties regex in `.clang-format` so it groups it properly.

Also, we can add `GSL/include` to `include_dirs` in `setup.py` so we can write `<gsl/span>` instead of `<GSL/include/gsl/span>`.

tests/test_clpir.py
```diff
@@ -744,5 +763,102 @@ def test_log_event_match(self) -> None:
...
+                            decoder_buffer = DecoderBuffer(
+                                initial_buffer_capacity=buffer_capacity, input_stream=istream
+                            )
+                        streaming_result = decoder_buffer._test_streaming(random_seed)
+                    except Exception as e:
+                        self.assertFalse(
+                            True, f"Error on file {file_path} using seed {random_seed}: {e}"
+                        )
+                self.__assert_streaming_result(file_path, streaming_result, random_seed)
+                random_seed += 1
```
```suggestion
```

src/clp_ffi_py/ir/PyDecoderBuffer.cpp
```diff
@@ -0,0 +1,313 @@
...
+    static char keyword_input_stream[]{"input_stream"};
+    static char keyword_initial_buffer_capacity[]{"initial_buffer_capacity"};
+    static char* keyword_table[]{
+            static_cast<char*>(keyword_input_stream),
+            static_cast<char*>(keyword_initial_buffer_capacity),
+            nullptr};
+
+    // If the argument parsing fails, `self` will be deallocated. We must reset
+    // all pointers to nullptr in advance, otherwise the deallocator might
+    // trigger segmentation fault
```
davidlion:
```suggestion
    // trigger a segmentation fault.
```

src/clp_ffi_py/ir/PyDecoderBuffer.hpp
```diff
@@ -0,0 +1,183 @@
...
+ * `IO[bytes]` like input stream. This class contains all the necessary data
+ * members to store the buffered bytes and track the buffer states. It is
+ * expected to be passed across different calls of CLP IR decoding methods when
+ * decoding from the same IR stream.
+ *
+ */
+class PyDecoderBuffer {
+    /**
+     * TODO: add class description to this class.
+     */
```
davidlion:
```suggestion
```

src/clp_ffi_py/ir/PyDecoderBuffer.hpp
```diff
@@ -0,0 +1,183 @@
...
+/**
+ * This Python Object represents a DecoderBuffer that buffers encoded CLP IR
+ * bytes read from the input stream. On the one hand, CLP IR decoding methods
+ * can consume the buffered bytes to decode log events. On the other hand, it
+ * follows the Python buffer protocol so that it can directly read from an
+ * `IO[bytes]` like input stream. This class contains all the necessary data
+ * members to store the buffered bytes and track the buffer states. It is
+ * expected to be passed across different calls of CLP IR decoding methods when
+ * decoding from the same IR stream.
+ *
```
davidlion, line 19:
```suggestion
```

src/clp_ffi_py/ir/PyDecoderBuffer.cpp
```diff
@@ -0,0 +1,313 @@
...
+        .bf_getbuffer = py_getbufferproc_cast(PyDecoderBuffer_getbuffer),
+        .bf_releasebuffer = py_releasebufferproc_cast(PyDecoderBuffer_releasebuffer),
+};
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays)
+PyDoc_STRVAR(
+        cPyDecoderBufferDoc,
+        "This class represents a CLP IR Decoder Buffer corresponding to a CLP IR stream. "
+        "It buffers encoded CLP IR data read from the input stream, which can be consumed by the "
+        "CLP IR decoding methods to recover encoded log events. The instance of this class is "
```
davidlion:
```suggestion
        "CLP IR decoding methods to recover encoded log events. An instance of this class is "
```

src/clp_ffi_py/ir/PyDecoderBuffer.hpp
```diff
@@ -0,0 +1,179 @@
...
+
+    /**
+     * Gets the PyTypeObject that represents PyDecoderBuffer's Python type. This
+     * type is dynamically created and initialized during the execution of
+     * `PyDecoderBuffer::module_level_init`.
+     * @return Python type object associated with PyDecoderBuffer.
+     */
+    [[nodiscard]] static auto get_py_type() -> PyTypeObject*;
+
+    /**:w
```
davidlion, line 144:
```suggestion
    /**
```

src/clp_ffi_py/ir/PyDecoderBuffer.hpp
```diff
@@ -0,0 +1,179 @@
...
+     *
```
davidlion:
```suggestion
    /**
```

src/clp_ffi_py/ir/PyDecoderBuffer.hpp
```diff
@@ -0,0 +1,179 @@
...
+     * This function should fail unless the buffer protocol is enabled.
+     * @param view Python Buffer view.
+     * @param flags Python Buffer flags.
+     * @return 0 on success.
+     * @return -1 on failure with the relevant Python exception and error set.
+     */
+    [[nodiscard]] auto py_getbuffer(Py_buffer* view, int flags) -> int;
+
+    /**
+     * @return Whether the buffer protocol is enabled.
```
davidlion:
```suggestion
     * @return true if the buffer protocol is enabled.
```

src/clp_ffi_py/ir/PyDecoderBuffer.hpp
```diff
@@ -0,0 +1,179 @@
...
+        m_buffer_size = 0;
+        m_num_current_bytes_consumed = 0;
+        m_num_decoded_message = 0;
+        m_py_buffer_protocol_enabled = false;
+        m_input_ir_stream = nullptr;
+    }
+
+    /**
+     * Releases the memory allocated for underlying metadata field and the
+     * reference hold for the Python object(s).
```
davidlion:
```suggestion
     * Releases the memory allocated for the underlying metadata field and the
     * reference held for the Python object(s).
```

src/clp_ffi_py/ir/PyDecoderBuffer.hpp
```diff
@@ -0,0 +1,178 @@
...
+     */
+    [[nodiscard]] auto is_py_buffer_protocol_enabled() const -> bool {
+        return m_py_buffer_protocol_enabled;
+    }
+
+    /**
+     * Tests the functionality of the DecoderBuffer by stepping through the
+     * input stream and stream it as random sized byte sequences. It will
+     * attempt to read a random number of bytes from the read buffer and grow
+     * the buffer when necessary until the the entire input stream is consumed.
```
davidlion:
```suggestion
     * Tests the functionality of the DecoderBuffer by sequentially reading
     * through the input stream with randomly sized reads. It will grow the read
     * buffer when necessary until the the entire input stream is consumed.
```

src/clp_ffi_py/ir/PyDecoderBuffer.hpp
LinZhihao-723: fixed

src/clp_ffi_py/ir/PyDecoderBuffer.cpp
LinZhihao-723: Resolved by using `gsl::span`

davidlion (approved):

PR #10, opened 2023-08-05
Implement IR decoding methods

LinZhihao-723:
# Description
Implement FFI IR decoding methods. These methods are wrapped into a Python class `clp_ffi_ir.Decoder` as static methods.
Since this change will complete most of the IR features (4-byte encoding) in the FFI library, a list of unit tests is added to test all the implemented methods as a whole system. These unit tests will first encode randomly generated log streams into the local file system using `FourByteEncoder`, and decode them using `Decoder`. In the meanwhile, the interactions between `Metadata`, `LogEvent`, `DecoderBuffer` and `Query` and the decoding methods are also tested. With the existing unit tests, we should have a more reliable testing system to ensure the FFI library has expected runtime behavior.

# Validation performed
1. All the existing unit tests passed.
2. Newly added unit tests passed as described above, which ensures the correctness of the entire encoding/decoding system.
3. Tested against the Uber benchmarks used in the beta-version code.
4. Passed all C++ and Python linters.


src/clp_ffi_py/ir/error_messages.hpp
```diff
@@ -3,6 +3,8 @@
 
 namespace clp_ffi_py::ir {
 constexpr char const* cDecoderBufferOverflowError = "DecoderBuffer internal read buffer overflows.";
+constexpr char const* cDecoderIncompleteIRError = "The IR stream is incomplete.";
+constexpr char const* cDecoderErrorCodeTemplate = "IR decoding method failed with error code: %d.";
```
davidlion:
```suggestion
constexpr char const* cDecoderErrorCodeFormatStr = "IR decoding method failed with error code: %d.";
```

Optional change:
I'd say calling this a format string is more common than a template.

tests/test_clpir.py
```diff
@@ -121,11 +110,187 @@ def __check_metadata(
...
+        )
+
+    def _check_wildcard_query(
+        self, wildcard_query: WildcardQuery, ref_wildcard_string: str, ref_is_case_sensitive: bool
+    ) -> None:
+        """
+        Given a WildcardQuery object, check if the stored data matches the input
+        reference.
+
+        :param self
```
```suggestion
```

tests/test_clpir.py
```diff
@@ -121,11 +110,187 @@ def __check_metadata(
...
+        ref_search_time_lower_bound: int,
+        ref_search_time_upper_bound: int,
+        ref_wildcard_queries: Optional[List[WildcardQuery]],
+        ref_search_time_termination_margin: int,
+    ) -> None:
+        """
+        Given a Query object, check if the stored data matches the input
+        references.
+
+        :param self
```
```suggestion
```

src/clp_ffi_py/ir/PyDecoder.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays)
+PyDoc_STRVAR(
+        cDecodeNextLogEventDoc,
+        "decode_next_log_event(decoder_buffer, query=None)\n"
+        "--\n\n"
+        "Decodes the next encoded log event from the IR stream buffered in the given decoder "
+        "buffer. `decode_preamble` must have been successfully invoked and returned with the "
+        "identical decoder buffer used in this argument. "
+        "If the query is provided, only the log event matching the query will be returned.\n"
```
davidlion:
```suggestion
        "buffer. `decoder_buffer` must have been returned by a successfully invocation of "
        "`decode_preamble`. If `query` is provided, only the next log event matching the query "
        "will be returned.\n"
```

src/clp_ffi_py/ir/decoding_methods.cpp
```diff
@@ -0,0 +1,250 @@
...
+    if (0 == num_bytes_read) {
+        PyErr_SetString(PyExc_RuntimeError, cDecoderIncompleteIRError);
+        return false;
+    }
+    return true;
+}
+
+/**
+ * Decodes the next log event that matches the query (if given) from the input
+ * DecoderBuffer using CLP decoding methods.
```
davidlion:
```suggestion
 * Decodes the next log event from the CLP IR buffer `decoder_buffer`. If
 * `py_query` is non-null decode until finding a log event that matches the
 * query.
```

src/clp_ffi_py/ir/decoding_methods.cpp
```diff
@@ -0,0 +1,250 @@
...
+namespace clp_ffi_py::ir {
+namespace {
+/**
+ * Attempts to populate the decoder buffer. When this function is called, it is
+ * expected to have more bytes to read from the IR stream.
+ * @param decoder_buffer Input decoder buffer.
+ * @return true on success.
+ * @return false on failure with the relevant Python exception and error set.
+ */
+auto try_read_more(PyDecoderBuffer* decoder_buffer) -> bool {
```
davidlion: Rename `try_read`.

davidlion (changes requested):
See comments.

tests/test_clpir.py
davidlion, line 1: Does this file still need to exist? I don't know all the interactions of unittest with main, but can it be omitted or moved somewhere else?

davidlion (approved):

PR #11, opened 2023-08-14
Adds a reader class as a wrapper to the decoding methods

LinZhihao-723:
# Description
Implements a simple reader class for users to use the decoding methods to iterate an encoded IR stream. It also provides a way to create a generator with a customized query. The overall structure and functionality should be similar to the reader implemented in the current CLP Python logging library.

# Validation performed
1. All the unit tests passed (including newly added ones)
2. All the linters passed.


src/clp_ffi_py/ir/decoding_methods.cpp
```diff
@@ -224,5 +224,75 @@ auto decode_next_log_event(PyObject* Py_UNUSED(self), PyObject* args, PyObject*
             is_query_given ? py_reinterpret_cast<PyQuery>(query) : nullptr
     );
 }
+
+auto skip_to_time(PyObject* Py_UNUSED(self), PyObject* args, PyObject* keywords) -> PyObject* {
```
davidlion: This method appears to duplicate a significant amount of code from `decode`. I think there should be a way to refactor the common code of these two methods.

clp_ffi_py/readers.py
```diff
@@ -0,0 +1,126 @@
...
+    generator with a customized search query.
+
+    :param istream: Input stream that contains encoded CLP IR.
+    :param decoder_buffer_size: Initial size of the decoder buffer.
+    :param enable_compression: A flag indicating whether the istream is
+    compressed using `zstd`.
+    """
+
+    def __init__(
+        self, istream: IO[bytes], decoder_buffer_size: int = 4096, enable_compression: bool = True
```
davidlion: Lets set the initial size of the buffer to something more practical. The initial value I used is pretty small relative to the logs we will process and cause a lot of reads.

tests/test_readers.py
```diff
@@ -0,0 +1,171 @@
...
+    def setUp(self) -> None:
+        self.enable_compression = False
+        self.has_query = True
+        self.num_test_iterations = 10
+        super().setUp()
+
+
+class TestCaseReaderTimeRangeQueryZst(TestCaseReaderTimeRangeQueryBase):
+    """
+    Tests stream reader against zst compressed IR stream with the query that
```
davidlion: zstd

tests/test_readers.py
```diff
@@ -0,0 +1,171 @@
...
+    def setUp(self) -> None:
+        self.enable_compression = False
+        self.has_query = True
+        self.num_test_iterations = 10
+        super().setUp()
+
+
+class TestCaseReaderWildcardQueryZst(TestCaseReaderWildcardQueryBase):
+    """
+    Tests stream reader against zst compressed IR stream with the query that
```
davidlion: zstd

tests/test_readers.py
davidlion, line 1: Zst and Zstd pop up interchangeably, we should probably use 'Zstd' everywhere.

LinZhihao-723: Everything should be "zstd" now to be consistent

davidlion (approved):

PR #14, opened 2023-08-22
Fix auto-completion issue resulting from class up-casting

jackluo923:
# Description
Auto-completion of `ClpIrStreamReader` is unavailable to IDEs due to down-casting of return type specified in the `__iter__` and `__enter__` class functions. This diff restores the auto-completion functionality by modifying `ClpIrStreamReader` to inherit the `Iterator` class, then switch the return type from `Iterator[LogEvent]` to `ClpIrStreamReader[LogEvent]` type. Additionally, `from __future__ import annotations` import statement is added to ensure lower Python runtime version can recognize the `ClpIrStreamReader` return type.

# Validation performed
After modifying `Reader.py` in the 0.0.2 release, the auto-completion functionality with IDE can be used normally.

### DIFF
--- a/clp_ffi_py/readers.py
+++ b/clp_ffi_py/readers.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from pathlib import Path
 from sys import stderr
 from types import TracebackType
@@ -8,7 +10,7 @@ from zstandard import ZstdDecompressionReader, ZstdDecompressor
 from clp_ffi_py import Decoder, DecoderBuffer, LogEvent, Metadata, Query
 
 
-class ClpIrStreamReader:
+class ClpIrStreamReader(Iterator):
     """
     This class represents a stream reader used to read/decode encoded log events
     from a CLP IR stream. It also provides method(s) to instantiate a log event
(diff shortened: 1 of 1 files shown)


clp_ffi_py/readers.py
```diff
@@ -8,7 +10,7 @@
 from clp_ffi_py import Decoder, DecoderBuffer, LogEvent, Metadata, Query
 
 
-class ClpIrStreamReader:
+class ClpIrStreamReader(Iterator):
```
LinZhihao-723:
```suggestion
class ClpIrStreamReader(Iterator[LogEvent]):
```

clp_ffi_py/readers.py
```diff
@@ -90,12 +92,12 @@ def search(self, query: Query) -> Generator[LogEvent, None, None]:
     def close(self) -> None:
         self.__istream.close()
 
-    def __iter__(self) -> Iterator[LogEvent]:
+    def __iter__(self) -> ClpIrStreamReader[LogEvent]:
```
LinZhihao-723:
```suggestion
    def __iter__(self) -> ClpIrStreamReader:
```

clp_ffi_py/readers.py
```diff
@@ -90,12 +92,12 @@ def search(self, query: Query) -> Generator[LogEvent, None, None]:
...
         if False is self.has_metadata():
             self.read_preamble()
         return self
 
-    def __enter__(self) -> Iterator[LogEvent]:
+    def __enter__(self) -> ClpIrStreamReader[LogEvent]:
```
LinZhihao-723:
```suggestion
    def __enter__(self) -> ClpIrStreamReader:
```

LinZhihao-723 (changes requested):
`Iterator` is a generic type, and it needs type annotation. We should annotate the inherited type when defining `ClpIrStreamReader`
Btw this should be caught by `mypy` :D

LinZhihao-723 (approved):

PR #16, opened 2023-08-22
Add example usage of `ClpIrStreamReader` class and slightly simplified the examples.

jackluo923:
# Description
Add example usage of `ClpIrStreamReader` class and slightly simplified the examples by removing the use of Pathlib library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,39 @@ from clp_ffi_py import Query
 help(Query)
 ```
 
-### Parallelization
+### Streaming Decode/Search Directly from S3 Remote Storage
+
+When working with CLP IR files stored on S3-compatible storage systems, [smart_open][17] can be used to
+open and read the IR stream for the following benefits:
+- It only performs stream operation and does not download the file to the disk.
+- It only invokes a single `GET` request so that the API access cost is minimized.
+
+Here is an example:
+
+```python
+from pathlib import Path
+from clp_ffi_py.readers import ClpIrStreamReader
+
+import boto3
+import os
+import smart_open
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -83,15 +83,16 @@ time_range_query: Query = Query(
 # A list to store all the log events within the search time range
 log_events: List[LogEvent] = []
 
-with ClpIrFileReader(Path("example.clp.zst")) as clp_reader:
-    for log_event in clp_reader.search(time_range_query):
-        log_events.append(log_event)
+# Open IRstream compressed log file as a binary file stream, then pass it to CLpIrStreamReader
```
LinZhihao-723:
```suggestion
# Open IRstream compressed log file as a binary file stream, then pass it to
# CLpIrStreamReader.
```
80-char limit

README.md
```diff
@@ -108,7 +109,8 @@ wildcard_search_query: Query = Query(wildcard_queries=wildcard_query_list)
 # [timestamp, message]
 matched_log_messages: List[Tuple[int, str]] = []
 
-with ClpIrFileReader(Path("example.clp.zst")) as clp_reader:
+# A convenience file reader class is also available to directly interact with IRstream compressed log file
```
LinZhihao-723:
```suggestion
# A convenience file reader class is also available to interact with a file that
# represents an encoded CLP IR stream directly.
```
80-char limit
Also we never used the term "compressed", so we should stick to the concept of "encoded" CLP IR stream

LinZhihao-723 (changes requested):

LinZhihao-723 (approved):

PR #18, opened 2023-08-23
Add allow_incomplete_stream support

LinZhihao-723:
# References
This PR adds a flag in the decoding method to allow truncated IR streams. If the flag is enabled, the truncated stream (incomplete stream error) will not raise any exceptions. Instead, it will be treated as the end of the stream.

# Description
Adds unit tests to validate the expected behavior.


src/clp_ffi_py/ir/PyFourByteEncoder.cpp
```diff
@@ -62,6 +62,15 @@ PyDoc_STRVAR(
         ":return: The encoded timestamp.\n"
 );
 
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays)
+PyDoc_STRVAR(
+        cEncodeEndOfIrDoc,
+        "encode_end_of_ir()\n"
+        "--\n\n"
+        "Encodes the EOIR that indicates the end of a CLP IR stream. A stream that does not "
+        "contain EOIR will be considered as an incomplete IR stream.\n"
```
davidlion: Might as well write out end of ir. Kinda weird to only abbreviate it here.

davidlion (approved):

davidlion (approved):

PR #27, opened 2023-09-28
Add support for partial matching in the wildcard query

LinZhihao-723:
# Description
In most cases, users want to search for any result containing the given wildcard search string. This is implemented as "partial matching", which is the default search query in `clg`. This PR implements partial matching as the default option when building the wildcard query.
In terms of the low level implementation details, a partial match wildcard query `"${WILDCARD_STRING}"` is equivalent to the full match wildcard query `*${WILDCARD_STRING}*`.

# Validation performed
1. Update the unit tests to cover the partial match.
2. All linters passed.

### DIFF
--- a/src/clp_ffi_py/ir/native/Query.hpp
+++ b/src/clp_ffi_py/ir/native/Query.hpp
@@ -25,37 +25,18 @@ public:
      * Initializes the wildcard query by cleaning the wildcard string.
      * @param wildcard_query Wildcard query.
      * @param case_sensitive Case sensitive indicator.
-     * @param partial_match Partial match indicator.
      */
-    WildcardQuery(std::string wildcard_query, bool case_sensitive, bool partial_match)
-            : m_original_query_string(std::move(wildcard_query)),
-              m_case_sensitive(case_sensitive),
-              m_partial_match(partial_match) {
-        if (partial_match) {
-            m_wildcard_query = "*";
-            m_wildcard_query += m_original_query_string;
-            m_wildcard_query += "*";
-            m_wildcard_query = clean_up_wildcard_search_string(m_wildcard_query);
-        } else {
-            m_wildcard_query = clean_up_wildcard_search_string(m_original_query_string);
-        }
-    }
-
-    [[nodiscard]] auto get_original_query_string() const -> std::string const& {
-        return m_original_query_string;
-    }
+    WildcardQuery(std::string wildcard_query, bool case_sensitive)
+            : m_wildcard_query(std::move(wildcard_query)),
+              m_case_sensitive(case_sensitive) {}
 
     [[nodiscard]] auto get_wildcard_query() const -> std::string const& { return m_wildcard_query; }
 
     [[nodiscard]] auto is_case_sensitive() const -> bool { return m_case_sensitive; }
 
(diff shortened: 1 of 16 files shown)


src/clp_ffi_py/ir/native/Query.hpp
```diff
@@ -8,33 +8,54 @@
...
+            m_wildcard_query = "*";
+            m_wildcard_query += m_uncleaned_wildcard_query;
+            m_wildcard_query += "*";
+            m_wildcard_query = clean_up_wildcard_search_string(m_wildcard_query);
+        } else {
+            m_wildcard_query = clean_up_wildcard_search_string(m_uncleaned_wildcard_query);
+        }
+    }
+
+    [[nodiscard]] auto get_uncleaned_wildcard_query() const -> std::string const& {
```
davidlion: Lets rename `get_uncleaned_wildcard_query`. Maybe something like `get_original_query_string` or `get_user_query_string`. I don't think we need to explicitly call it a wildcard query considering this is a method of the `WildcardQuery` class (so pretty redundant). I'm a bit indifferent about adding `_string` as a suffix, but it would probably be necessary if there is ever a `Query` class in the future.

LinZhihao-723: Agree. I will keep the suffix `_string` to differentiate it from the actual query and the input.

junhaoliao:
As we discussed offline, without partial matching, the current `.search()` behaviour has caused confusion among users. Shall we revive this PR?

LinZhihao-723:
> As we discussed offline, without partial matching, the current `.search()` behaviour has caused confusion among users. Shall we revive this PR?

@kirkrodrigues I think we have a clear use case where users want the wildcard to be automatically surrounded by `*`

kirkrodrigues:
> > As we discussed offline, without partial matching, the current `.search()` behaviour has caused confusion among users. Shall we revive this PR?
> 
> @kirkrodrigues I think we have a clear use case where users want the wildcard to be automatically surrounded by `*`

Things are a little awkward because users are already using this library in production and the recommended practice is not to change APIs without a major version bump; to support new functionality before the major version bump, the recommendation is to mark the current API for deprecation and then add new APIs with the new functionality.

In light of that, my proposal would be:

* Add two new classes---`SubstringWildcardQuery` and `FullStringWildcardQuery`---that are derived from `WildcardQuery`.
  * I prefer this compared to a flag since it makes the scope of the query clear.
  * Internally, `SubstringWildcardQuery` can call the `WildcardQuery` constructor with the added prefix & suffix wildcards. Eventually, if we need to move the prefix/suffix logic lower in the call hierarchy (e.g., for debugging as you suggested initially), then having a separate class makes this simpler.
* Mark `WildcardQuery` for deprecation. In the future, we can make it abstract like @davidlion suggested when we discussed this initially.
* Mark `QueryBuilder`'s `add_wildcard_query` for deprecation and add a new version of `add_wildcard_query` (using Python's syntactic sugar for method overloading) which takes a `WildcardQuery` instead.
  * One advantage of this new `add_wildcard_query` is we don't have to keep changing the API of `add_wildcard_query` whenever the `WildcardQuery` constructors change. It's also more consistent with `add_wildcard_queries`.

LinZhihao-723:
The equivalence has been implemented in PR #62

PR #28, opened 2023-10-26
Fix: implement trivial destructors for static PyObject unique pointers to avoid invalid memory access on exit

LinZhihao-723:
# Description
The static destructors of the global/static PyObject unique pointers (such as Py types and exceptions) will attempt to decrease the reference count when the process exits. However, its behavior is undefined due to the following reasons, which was first exposed in Python 3.10.12 on Linux with a segfault on exit:

1. the Python interpreter can exit before the global/static destructors execute. Accessing the reference count is illegal if the interpreter is already shut down. There is no way to guarantee the ordering of the execution for the interpreter cleanup and the global/static destructors.
2. C++ RAII guarantees that some code will execute when the scope exits, but it doesn't guarantee that all of the code finishes executing; the process may exit before the RAII destructor is complete.

This PR fixes the problem by making all the global/static Pyobject pointers trivially destructible. This means it will not attempt to release any resources when reset or destruction happens, which can lead to potential memory leaks. However, according to the current use case, this is acceptable since the only expected execution of the trivial destructor is on process exit. 

We keep using smart ptr for two purposes: 
1. protect the underlying raw pointers from being exposed and accidentally modified.
2. indicate the underlying object will never be freed by C++ static destructor.

### DIFF
--- a/src/clp_ffi_py/PyObjectUtils.hpp
+++ b/src/clp_ffi_py/PyObjectUtils.hpp
@@ -38,10 +38,14 @@ template <typename PyObjectType>
 using PyObjectPtr = std::unique_ptr<PyObjectType, PyObjectDeleter<PyObjectType>>;
 
 /**
- * A type of smart pointer that holds a static Python object raw pointer. For
- * static/global variables, the destructor will be executed after the Python
- * interpreter exits. Therefore, the destructor is set to empty to avoid unsafe
- * memory operations.
+ * A smart pointer to be used for raw Python object pointers that have static
+ * storage duration. It holds a reference of the underlying Python object.
+ * Compared to PyObjectPtr, when destructed, this smart pointer does not
+ * decrease the reference count of the contained Python object. Since this
+ * pointer has static storage duration, it's possible that the Python
+ * interpreter exits before this pointer's destructor is called; attempting to
+ * decrement the reference count (maintained by the interpreter) in this
+ * situation would lead to undefined behaviour.
  * @tparam PyObjectType
  */
 template <typename PyObjectType>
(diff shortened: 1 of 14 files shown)


src/clp_ffi_py/Py_utils.cpp
```diff
@@ -7,10 +7,10 @@
 namespace clp_ffi_py {
 namespace {
 constexpr char const* const cPyFuncNameGetFormattedTimestamp{"get_formatted_timestamp"};
-PyObjectPtr<PyObject> Py_func_get_formatted_timestamp;
+PyObjectStaticPtr<PyObject> Py_func_get_formatted_timestamp;
 
 constexpr char const* const cPyFuncNameGetTimezoneFromTimezoneId{"get_timezone_from_timezone_id"};
-PyObjectPtr<PyObject> Py_func_get_timezone_from_timezone_id;
+PyObjectStaticPtr<PyObject> Py_func_get_timezone_from_timezone_id;
```
davidlion: Lets initialize these to `nullptr` like all the others.

src/clp_ffi_py/PyObjectUtils.hpp
```diff
@@ -18,12 +18,33 @@ class PyObjectDeleter {
...
+
+/**
+ * A type of smart pointer that holds a static Python object raw pointer. For
+ * static/global variables, the destructor will be executed after the Python
+ * interpreter exits. Therefore, the destructor is set to empty to avoid unsafe
+ * memory operations.
+ * @tparam PyObjectType
+ */
+template <typename PyObjectType>
+using PyObjectStaticPtr = std::unique_ptr<PyObjectType, PyObjectTrivialDeleter<PyObjectType>>;
```
davidlion: This description is good, but I think we can add more detail from the PR here.

davidlion (changes requested):

davidlion (approved):

PR #33, opened 2023-11-11
Update setup.py to explicitly include submodules and their files when building for Python3.6

LinZhihao-723:
# Description
When building for Python3.6, files from the submodules are not recursively included, and we need to explicitly include `.py` and `.typed` to ensure they are a part of the submodules. This PR updates the `setup.py` to properly handle the Python3.6 building process.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ if "__main__" == __name__:
 
         project_name: str = "clp_ffi_py"
         description: str = "CLP FFI Python Interface"
-        extension_modules: List[Exception] = [ir_native]
+        extension_modules: List[Extension] = [ir_native]
         if (3, 7) > sys.version_info:
             # For Python3.6, setuptools doesn't automatically include submodules
             # and .pyi/.type files, so we need to explicitly specify the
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -55,13 +55,29 @@
...
+            # For Python3.6, we need to explicitly specify the packages and the
+            # package data. Submodules and .pyi/.type files are not
+            # automatically included.
+            packages: List[str] = ["clp_ffi_py", "clp_ffi_py.ir"]
+            data_to_include: List[str] = ["*.py", "*.pyi", "*.typed"]
+            package_data: Dict[str, List[str]] = {package: data_to_include for package in packages}
+            setup(
+                name="clp_ffi_py",
+                description="CLP FFI Python Interface",
+                ext_modules=[ir_native],
```
kirkrodrigues: Can we extract these into variables so that they don't go out-of-sync between the `if` and `else`?

kirkrodrigues (changes requested):

setup.py
LinZhihao-723: Yeah, make sense, will do it

kirkrodrigues (approved):

PR #37, opened 2023-11-26
Replace memcpy with std::copy to avoid undefined behaviour when buffers overlap.

LinZhihao-723:
# Description
In the current implementation, the code uses `memcpy` to copy the unconsumed bytes from the end of the buffer to the beginning. In most cases, this is safe because the previous `if` statement handles the cases where more than half of bytes in the current buffer are unconsumed. However, the corner case happens when the number of unconsumed bytes is larger than half of the actual buffer size (not the capacity). If this happens, the behavior of the code depends on the underlying `memcpy` implementation. This PR solves this UB by replacing `memcpy` with `std::copy`, so it will be safe even if the src and dst buffers overlap.
Notice that there will be a performance drop with the changes in this PR since `std::copy` cannot optimize `gsl::span`. However, this problem will be fix after switching to C++20 with `std::span`.


davidlion (approved):

davidlion (changes requested):
Lets make this copy use c++.

davidlion (approved):

PR #39, opened 2023-11-29
Upgrade the C++ standard to C++20; Replace gsl::span with std::span.

LinZhihao-723:
# Description
This PR upgrades the C++ standard from C++17 to C++20. With C++20 enabled, gsl::span can be now replaced by std::span.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,11 +37,13 @@ color = true
 preview = true
 
 [tool.cibuildwheel]
+# Use `musllinux_1_2` to support C++20 compilation
 musllinux-aarch64-image = "musllinux_1_2"
 musllinux-i686-image = "musllinux_1_2"
 musllinux-ppc64le-image = "musllinux_1_2"
 musllinux-s390x-image = "musllinux_1_2"
 musllinux-x86_64-image = "musllinux_1_2"
+
 skip = "pp*"
 
 test-command = [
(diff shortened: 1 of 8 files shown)


src/clp_ffi_py/ir/native/PyDecoderBuffer.cpp
```diff
@@ -241,7 +242,7 @@ auto PyDecoderBuffer::populate_read_buffer(Py_ssize_t& num_bytes_read) -> bool {
             PyErr_NoMemory();
             return false;
         }
-        auto new_read_buffer{gsl::span<int8_t>(new_buf, new_capacity)};
+        auto new_read_buffer{std::span<int8_t>(new_buf, new_capacity)};
```
kirkrodrigues: Are you writing `auto x = {Constructor(...)}` to avoid the explicit cast of `new_capacity`?

pyproject.toml
```diff
@@ -37,6 +37,11 @@ color = true
 preview = true
 
 [tool.cibuildwheel]
+musllinux-aarch64-image = "musllinux_1_2"
```
kirkrodrigues, line 41: This is optional, but it might be worth adding a comment about why we're using musllinux_1_2.

kirkrodrigues (changes requested):
We should also remove the gsl submodule, right?

LinZhihao-723:
Didn't remove GSL since it seems useful for future changes. But sure, we can add it back when needed.

kirkrodrigues (approved):

PR #41, opened 2023-12-04
Add support to skip forward when decoding IR streams

LinZhihao-723:
# Description
Currently, the decoder doesn't support skipping forward. Users have to decode and discard log events when needed manually. This PR enables forward skipping by introducing a new method to decode and discard a certain amount of log events from the given IR stream.
This PR also refactors the implementation of the decoding method to template the termination of a decoding process. This provides a function interface to reuse decoding logic and implement query search decoding, sequential decoding, and forward skipping.

# Validation performed
1. Add new unit tests to verify the behavior of forward skipping.
2. Ensure all existing unit tests are passed after refactoring the decoding method.

### DIFF
--- a/clp_ffi_py/ir/native.pyi
+++ b/clp_ffi_py/ir/native.pyi
@@ -79,5 +79,11 @@ class Decoder:
         query: Optional[Query] = None,
         allow_incomplete_stream: bool = False,
     ) -> Optional[LogEvent]: ...
+    @staticmethod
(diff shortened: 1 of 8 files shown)


kirkrodrigues:
Remind me, what was the conclusion we reached for this PR?

LinZhihao-723:
This PR should be no longer needed, as it's functionality should be handled by the query in nature

PR #42, opened 2023-12-04
Refactor decoding method implementation to generalize decoding termination logic.

LinZhihao-723:
# Description
This PR refactors the decoding method's implementation to template the decoding process's termination. It provides a function interface to reuse decoding logic and implement query search decoding and sequential decoding as lambda termination handlers.

### DIFF
--- a/src/clp_ffi_py/ir/native/decoding_methods.cpp
+++ b/src/clp_ffi_py/ir/native/decoding_methods.cpp
@@ -23,16 +23,30 @@
 namespace clp_ffi_py::ir::native {
 namespace {
 /**
+ * This template defines the function signature of a termination handler
+ * required by `generic_decode_log_event`. Signature: (
+ *         ffi::epoch_timestamp_ms timestamp,
+ *         std::string_view decoded_log_message,
+ *         size_t decoded_log_event_idx,
+ *         PyObject*& return_value
+ * ) -> bool;
+ * @tparam TerminateHandler
+ */
+template <typename TerminateHandler>
+concept TerminateHandlerSignature = requires(TerminateHandler handler) {
+    {
+        handler(std::declval<ffi::epoch_time_ms_t>(),
+                std::declval<std::string_view>(),
+                std::declval<size_t>(),
+                std::declval<PyObject*&>())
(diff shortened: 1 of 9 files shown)


src/clp_ffi_py/ir/native/decoding_methods.cpp
```diff
@@ -23,30 +23,36 @@
...
+ * Decodes the next log event from the CLP IR buffer `decoder_buffer` until
+ * terminate handler returns true.
+ * @tparam TerminateHandler Method to determine if the decoding should
+ * terminate, and set the return value for termination.
+ * Signature: (
+ *         [maybe_unused] ffi::epoch_timestamp_ms timestamp,
+ *         [maybe_unused] std::string_view decoded_log_message,
+ *         [maybe_unused] size_t decoded_log_event_idx,
+ *         PyObject*& return_value
+ * ) -> bool;
```
davidlion: Why is TerminateHandler a template type and not a normal function pointer param?

src/clp_ffi_py/ir/native/decoding_methods.cpp
```diff
@@ -59,22 +65,23 @@ auto decode(
...
+            if (false == is_incomplete_stream_error) {
+                return nullptr;
+            }
+            PyErr_Clear();
+            Py_RETURN_NONE;
         }
         if (ffi::ir_stream::IRErrorCode_Eof == err) {
-            reached_eof = true;
-            break;
+            Py_RETURN_NONE;
```
davidlion: Feels as though this error logic could be made more clear (maybe with just variable naming).

davidlion (changes requested):
See comments.

davidlion (approved):

src/clp_ffi_py/ir/native/decoding_methods.cpp
LinZhihao-723: Here's a conclusion of our off-line discussion:
1. We should prefer lambda rather than C style function pointer or `std::function` since it is more lightweight.
2. We should use `concept` (C++20) to specify the function signature.

PR #47, opened 2024-03-15
Lock smart_open to v6.4.0 to fix unit-test and Python3.6 build failures.

LinZhihao-723:
# References

# Description

`smart-open` has released version 7.0.1. In this new version, they introduced `__future__` notations, which were not supported in Python3.6, thus causing the unit test failures. Also, it attempts to use the `_FileLikeProxy__inner` attributes of `zstd.ZstdCompressionWriter`, which does not exist. Therefore, we will lock `smart-open` to an earlier version to avoid unit test failures.

# Validation performed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,7 @@ test-command = [
   "python -m unittest -fv",
 ]
 test-requires = [
-    "smart_open"
+    "smart_open==6.4.0"
 ]
 test-skip = [
     "cp37-*",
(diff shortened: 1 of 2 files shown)


kirkrodrigues (approved):
How about:

Lock smart_open to v6.4.0 to fix unit-test and Python3.6 build failures. (#47)

LinZhihao-723:
> How about:
> 
> Lock smart_open to v6.4.0 to fix unit-test and Python3.6 build failures. (#47)

LGTM

PR #46, opened 2024-03-14
Add Taskfile to build package; Refactor docs tasks.

kirkrodrigues:
# Description

This PR:
* Adds a [Taskfile](https://taskfile.dev/usage/) to make it easier to build clp_ffi_py
* Refactors the docs Taskfile so that it can be called from the main Taskfile.
* Applies our latest guidelines for Taskfiles to the docs tasks.

NOTE: This PR requires task v3.35.1 for the latest bug fixes.

# Validation performed

* Validated building each artifact from a clean state:

  ```bash
  task clean && task
  task clean && task wheel
  task clean && task docs:site
  ```

* Validated repeated builds of each artifact-generating task did not result in the artifact being regenerated:

  ```bash
  task clean && task wheel && task wheel
  task clean && task docs:site && task docs:site
  ```

* Validated that after running `task docs:site`, changing any of the following files triggered a rebuild of just the docs:
  * `docs/conf/conf.py`
  * `docs/src/index.md`
  * `docs/tasks.yml`
* Validated that after running `task wheel`, changing any of the following triggered a rebuild of the wheel:
  * `clp_ffi_py/__init__.py`
  * `src/clp` (the version of the clp submodule)
  * `src/clp_ffi_py/error_messages.hpp`
  * `Taskfile.yml`
* Validated that after running `task docs:site`, changing any of the following files from the previous step triggered a rebuild of the wheel and the docs.
* Validated `task clean && task docs:serve` built and served the site.

PR #48, opened 2024-03-15
Commit API docs and generate them using a separate task.

kirkrodrigues:
# Description

Currently, clp-ffi-py's API docs are built whenever the docs site is built. This has two issues:

* We can't apply any manual customizations to the API docs, if we wanted to.
* The edit links in the docs site won't work for the API docs since they aren't committed to the repo.

This PR commits the API docs to the repo and builds them using a separate task that devs should run manually if they ever add/remove a package/module.

This PR depends on #46 being merged first.

# Validation performed

* Validated `task docs:serve` continues to work correctly.
* Validated `task docs:api-docs` generates the API docs but doesn't change the committed docs (right now).


docs/README.md
```diff
@@ -21,14 +21,29 @@ documentation website for this package:
...
   task docs:clean
   ```
 
+* Generate API docs:
+
+  ```shell
+  task docs:api-docs
+  ```
+  
+  * This will regenerate the API docs in `src/api`. You should run this whenever you change what
```
haiqi96: do we need the "what" in the sentence?

docs/src/api/clp_ffi_py.ir.native.rst
```diff
@@ -0,0 +1,7 @@
+clp\_ffi\_py.ir.native module
```
haiqi96, line 1: I have no idea what are those so I assume they are auto generated

haiqi96 (commented):
Only a small comment on the grammer

docs/src/api/clp_ffi_py.ir.native.rst
kirkrodrigues: Yessir.

docs/README.md
kirkrodrigues: Updated the wording.

haiqi96 (approved):

PR #51, opened 2024-03-15
gh-actions: Parallelize wheel-building by explicitly specifying wheels to build in job matrix.

kirkrodrigues:
# Description

This repo's workflows for pull requests currently take about 2 hours to run making it tedious to merge any PRs. This is because we use a single job to build all the wheels for Linux and another to build all the wheels for macOS.

This PR changes the workflow to use a job to build each wheel. It does this by specifying the wheels to build individually.

This drops the workflow time down to around 15 minutes (8x improvement). Of course, it relies on GH having enough spare compute capacity, so in the worst case, the workflows could still end up taking 2 hours, but at least it's not the average case.

NOTE: Since each wheel is now built in a separate job, we upload each wheel as a separate artifact.

This PR also updates all actions to the latest versions to fix some deprecation warnings.

# Validation performed

Validated the workflows ran successfully.


kirkrodrigues:
@LinZhihao-723, since each wheel is now a separate upload, does it significantly impact the release process?

LinZhihao-723:
1. For the release process, we didn't have it automated yet. I don't see any problem here.
2. I don't think we should touch `pyproject.toml` since the settings there would still be useful when building the wheels locally.

LinZhihao-723:
Also, is it possible to collect the generated artifacts into one archive?

pyproject.toml
```diff
@@ -60,8 +64,12 @@ test-skip = [
     "*-*linux_i686",
 ]
 
+[tool.cibuildwheel.linux]
+archs = ["auto", "aarch64"]
+
 [tool.cibuildwheel.macos]
 archs = ["x86_64", "universal2", "arm64"]
+environment = { MACOSX_DEPLOYMENT_TARGET = "10.15" }
```
LinZhihao-723: We don't need this. `setup.py` will set this env variable when needed, which ensures the build process to work without using cibuildwheel.

LinZhihao-723 (changes requested):

pyproject.toml
kirkrodrigues: Do we need it in the workflow then?

LinZhihao-723: I think we don't need to explicitly specify it in the workflow

kirkrodrigues: Looks like we do: https://github.com/y-scope/clp-ffi-py/actions/runs/8307258601/job/22736272998.

Do you have a solution or should we add it back to `pyproject.toml`?

LinZhihao-723: Sorry my bad. Here's the solution: https://github.com/y-scope/clp-ffi-py/pull/52

LinZhihao-723 (approved):

LinZhihao-723:
Commit message lgtm

PR #52, opened 2024-03-16
Make macOS deployment target check robust against float precision errors.

LinZhihao-723:
# References

# Description

When building for MacOS, we used to compare the version by converting the version to a floating point, which didn't work in `cibuildwheel` environment. This PR properly extracts the major and minor versions and executes int comparison instead. By doing this, we no longer need to specify the env variable in the workflow. This will ensure the local build works in MacOS environment.

# Validation performed

Validated workflow passed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,14 +45,14 @@ if "__main__" == __name__:
     try:
         if "Darwin" == platform.system():
             target_env_var_name: str = "MACOSX_DEPLOYMENT_TARGET"
-            required_target: Tuple[int, int] = (10, 15)
+            min_target_version: Tuple[int, int] = (10, 15)
             
             target_str: Optional[str] = os.environ.get(target_env_var_name)
             target: Tuple[int, ...] = ()
             if target_str is not None:
                 target = tuple(int(part) for part in target_str.split("."))
-            if target < required_target:
-                target = required_target
+            if target < min_target_version:
+                target = min_target_version
             os.environ[target_env_var_name] = ".".join(str(part) for part in target)
 
         project_name: str = "clp_ffi_py"
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ import os
 import platform
 import sys
 from setuptools import setup, Extension
-from typing import List, Optional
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -45,14 +45,21 @@
...
-            if None is target or float(target) < 10.15:
+            env_setup_needed: bool = False
+            if None is target:
+                env_setup_needed = True
+            else:
+                version_values: List[int] = [int(part) for part in target.split(".")]
+                if 10 >= version_values[0] and 15 >= version_values[1]:
+                    env_setup_needed = True
+            if env_setup_needed:
                 os.environ["MACOSX_DEPLOYMENT_TARGET"] = "10.15"
```
kirkrodrigues: How about something like:

```suggestion
            target_env_var_name = "MACOSX_DEPLOYMENT_TARGET"
            required_target = (10, 15)
            
            target_str: Optional[str] = os.environ.get(target_env_var_name)
            target: Tuple[int, ...] = ()
            if target_str is not None:
                target = tuple(int(part) for part in target_str.split("."))
            if target < required_target:
                target = required_target
            os.environ[target_env_var_name] = ".".join(str(part) for part in target)
```

That way we aren't repeating any constants (besides the version separator) and the version comparison is more Pythonic.

kirkrodrigues (changes requested):

setup.py
LinZhihao-723: Sure. Also added type annotations for `target_env_var_name` and `required_target `

kirkrodrigues (approved):
How about:

Make macOS deployment target check robust against float precision errors. (#52)

setup.py
kirkrodrigues: Maybe `required_target` should be `min_target`.

LinZhihao-723: Sure

LinZhihao-723:
> How about:
> 
> Make macOS deployment target check robust against float precision errors. (#52)

lgtm

kirkrodrigues (approved):

PR #53, opened 2024-03-16
Add support for compiling C-extension in parallel.

LinZhihao-723:
# References

# Description

The native `setup.py` support doesn't provide a way to parallelize source compilation, making the compilation time very slow. This PR implements a patch to enable the parallel compilation by learning [grpc](https://github.com/grpc/grpc/blob/master/src/python/grpcio/_parallel_compile_patch.py) and [cpython](https://github.com/python/cpython/blob/31368a4f0e531c19affe2a1becd25fc316bc7501/Lib/distutils/ccompiler.py#L511C9-L511C16). Essentially, this PR implements a method that overloads the default `compile` method to use `multiprocessing` to compile files in parallel.

Local Benchmark:
When building for MacOS in M1 Pro (10 cores), the total compilation time (real-time) using `pip install -e .` is:
Serialized compilation (before this PR): 39.365s
Parallel compilation (after this PR): 11.709s
Improvements: 3.362x

Github Action Benchmark:
Unfortunately, Github Action only provides 2 cores in their virtual environment, so the improvements are not obvious. However, we can still see the improvements by comparing these two actions:
Serialized compilation (before this PR): https://github.com/y-scope/clp-ffi-py/actions/runs/8307187935
Parallel compilation (after this PR): https://github.com/LinZhihao-723/clp-ffi-py/actions/runs/8308605670
We may cherry-pick some builds to check the wheel building time:
cp38-manylinux_aarch64: 386.13s -> 223.95s
cp310-manylinux_i686: 41.50s -> 20.43s
cp311-macosx_arm64: 51.07s -> 22.02s
Note: the time in macos build doesn't have a stable improvement, which might be because the virtualized macos env doesn't have 2 physical cores.

# Validation performed

Validated the workflow still passed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,10 @@
+import distutils.ccompiler
+import multiprocessing.pool
 import os
 import platform
 import sys
 from setuptools import setup, Extension
-from typing import List, Optional
+from typing import List, Optional, Tuple
 
 ir_native: Extension = Extension(
     name="clp_ffi_py.ir.native",
@@ -41,6 +43,33 @@ ir_native: Extension = Extension(
     define_macros=[("SOURCE_PATH_SIZE", str(len(os.path.abspath("./src/clp/components/core"))))],
 )
 
+def _parallel_compile(
+    self: distutils.ccompiler.CCompiler,
+    sources: List[str],
+    output_dir: Optional[str] = None,
+    macros: Optional[List[Tuple[str, Optional[str]]]] = None,
+    include_dirs: Optional[List[str]] = None,
+    debug: int = 0,
+    extra_preargs: Optional[List[str]] = None,
+    extra_postargs: Optional[List[str]] = None,
+    depends: Optional[List[str]] = None,
+) -> List[str]:
+    macros, objects, extra_postargs, pp_opts, build = self._setup_compile(
+        output_dir, macros, include_dirs, sources, depends, extra_postargs
+    )
+    cc_args = self._get_cc_args(pp_opts, debug, extra_preargs)
+
+    def _compile_single_file(obj: str) -> None:
+        try:
+            src, ext = build[obj]
+        except KeyError:
+            return
+        self._compile(obj, src, ext, cc_args, extra_postargs, pp_opts)
+
+    num_cores: int = multiprocessing.cpu_count()
+    multiprocessing.pool.ThreadPool(num_cores).map(_compile_single_file, objects)
+    return objects
+
 if "__main__" == __name__:
     try:
         if "Darwin" == platform.system():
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -41,13 +43,41 @@
...
+    sources: List[str],
+    output_dir: Optional[str] = None,
+    macros: Optional[List[Tuple[str, Optional[str]]]] = None,
+    include_dirs: Optional[List[str]] = None,
+    debug: int = 0,
+    extra_preargs: Optional[List[str]] = None,
+    extra_postargs: Optional[List[str]] = None,
+    depends: Optional[List[str]] = None,
+) -> List[str]:
+    macros, objects, extra_postargs, pp_opts, build = self._setup_compile(
```
kirkrodrigues, line 61:
```suggestion
    """
    A replacement for distutils.ccompiler.CCompiler.compile that compiles each source file
    concurrently. NOTE: This relies on building the wheel using Python <= 3.11.
    """
    macros, objects, extra_postargs, pp_opts, build = self._setup_compile(
```

setup.py
```diff
@@ -41,13 +43,41 @@
...
+
+    def _compile_single_file(obj: str) -> None:
+        try:
+            src, ext = build[obj]
+        except KeyError:
+            return
+        self._compile(obj, src, ext, cc_args, extra_postargs, pp_opts)
+
+    num_cores: int = multiprocessing.cpu_count()
+    multiprocessing.pool.ThreadPool(num_cores).map(_compile_single_file, objects)
```
kirkrodrigues: We should use `ThreadPool` with `with`, right (the docs say it should be [closed](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.ThreadPool))?

kirkrodrigues (changes requested):

setup.py
LinZhihao-723: right, mb

kirkrodrigues (approved):
How about:

Add support for compiling C-extension using all cores. (#53)

setup.py
LinZhihao-723: At this moment we are still able to build the wheels using Python3.12. I will drop the last sentence.

LinZhihao-723:
> How about:
> 
> Add support for compiling C-extension using all cores. (#53)

Shall we use "Add support for compiling C-extension using all cores **in parallel**." or "Add support for compiling C-extension using all cores in **concurrently**."? I think we should be explicit about the parallelism using some key words.

kirkrodrigues:
Sure, we can use your first suggestion or this one:

Add support for compiling C-extension in parallel. (#53)

kirkrodrigues (approved):

LinZhihao-723:
> Sure, we can use your first suggestion or this one:
> 
> Add support for compiling C-extension in parallel. (#53)

lgtm

PR #62, opened 2024-06-03
Add more explicit interface to differentiate substring and full-string wildcard queries.

LinZhihao-723:
# References

# Description

As discussed in PR #27, users usually want the query to be automatically surrounded with wildcard `*` to match a substring. This PR implements such a query in a way we discussed and agreed on in PR #27. Essentially, this PR contains the following changes:
- Redesign `WildcardQuery` classes. `WildcardQuery` is now marked as deprecated. Instead, users should use its derived class `SubstringWildcardQuery` and `FullStringWildcardQuery` to define a wildcard query. When using `SubstringWildcardQuery`, `*` prefix and postfix will be automatically added to match a substring.
- Redesign `QueryBuilder` to change the method's signature of `add_wildcard_query`. It's taking a `WildcardQuery` class instead of a string and a boolean argument. The old signature can still be called but it is marked as deprecated.

We can support substring matches without breaking existing user code with the above implementations. In the future, we should also do the following cleanup:
- Formally remove the old signature of `add_wildcard_query`
- Make `WildcardQuery` as an abstract base class
- Close PR #27

# Validation performed

- All unit tests passed
- Add unit tests to cover deprecated functions
- CI passed, successfully built and tested across all the supported platforms and Python versions

### DIFF
--- a/clp_ffi_py/ir/query_builder.py
+++ b/clp_ffi_py/ir/query_builder.py
@@ -117,9 +117,6 @@ class QueryBuilder:
     ) -> QueryBuilder:
         """
         This method is the implementation of `add_wildcard_query`.
-
-        Type check is disabled since it executes runtime checks to ensure
-        passed-in arguments match the defined signatures.
         """
         if isinstance(wildcard_query, WildcardQuery):
             self._wildcard_queries.append(wildcard_query)
--- a/clp_ffi_py/ir/query_builder.py
+++ b/clp_ffi_py/ir/query_builder.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import warnings
 from copy import deepcopy
 from typing import Any, Dict, List, no_type_check, Optional, overload, Tuple, Union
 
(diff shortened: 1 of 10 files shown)


clp_ffi_py/ir/query_builder.py
```diff
@@ -87,8 +99,61 @@ def add_wildcard_query(self, wildcard_query: str, case_sensitive: bool = False)
...
+        Adds the given wildcard query to the wildcard query list.
+
+        :param wildcard_query: The wildcard query to add. It can be any derived
+            class of :class:`~clp_ffi_py.wildcard_query.WildcardQuery`.
+        :return: self.
+        """
+        ...
+
+    @no_type_check
+    def add_wildcard_query(self, *args: Tuple[Any, ...], **kwargs: Dict[str, Any]) -> QueryBuilder:
```
davidlion: As discussed lets drop `no_type_check` and lets try a different approach.

davidlion (changes requested):
Improve `add_wildcard_query`

davidlion (approved):
Lets change the title + commit message to something more direct like:
`Add more explicit interface to differentiate substring and full-string wildcard queries. (#62).`

LinZhihao-723:
> Lets change the title + commit message to something more direct like: `Add more explicit interface to differentiate substring and full-string wildcard queries. (#62).`

The commit msg lgtm, thx!

PR #63, opened 2024-06-12
Docs: Clean-up docstrings for wildcard query methods and classes introduced in #62.

LinZhihao-723:
# References

# Description

The docstrings in the new methods and classes introduced in #62 were not properly rendered in the doc webpage. There are two problems to solve:
1. For overloaded methods, no auto docstring generation is supported: https://github.com/sphinx-doc/sphinx/issues/7787. Therefore, we need to manually write docstrings for every signature in Sphinx documentation format.
2. We need to add a docstring for `FullStringWildcardQuery`'s init function, otherwise, Sphinx will use the base class init, which is marked as `deprecated` in this case.

# Validation performed

Build the site locally and ensure it is rendered as expected.
1. For `add_wildcard_query`:
image

2. For `FullStringWildcardQuery`:
image

### DIFF
--- a/clp_ffi_py/wildcard_query.py
+++ b/clp_ffi_py/wildcard_query.py
@@ -19,10 +19,10 @@ class WildcardQuery:
 
     @deprecated(
         version="0.0.12",
-        reason="`clp_ffi_py.wildcard_query.WildcardQuery` is supposed to be an abstract class and"
-        " should not be used directly. To create a wildcard query, please explicit instantiate"
-        " `clp_ffi_py.wildcard_query.SubstringWildcardQuery` or"
-        " `clp_ffi_py.wildcard_query.FullStringWildcardQuery`.",
+        reason=":class:`~clp_ffi_py.wildcard_query.WildcardQuery` is supposed to be an abstract"
+        " class and should not be used directly. To create a wildcard query, please explicit"
+        " instantiate :class:`~clp_ffi_py.wildcard_query.SubstringWildcardQuery` or"
+        " :class:`~clp_ffi_py.wildcard_query.FullStringWildcardQuery`.",
     )
     def __init__(self, wildcard_query: str, case_sensitive: bool = False):
         """
--- a/clp_ffi_py/wildcard_query.py
+++ b/clp_ffi_py/wildcard_query.py
@@ -92,6 +92,12 @@ class FullStringWildcardQuery(WildcardQuery):
     """
 
     def __init__(self, full_string_wildcard_query: str, case_sensitive: bool = False):
+        """
+        Initializes a full-string wildcard query using the given parameters.
+
+        :param full_string_wildcard_query: Wildcard query string.
+        :param case_sensitive: Case sensitive indicator.
+        """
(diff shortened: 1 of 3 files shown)


clp_ffi_py/wildcard_query.py
```diff
@@ -19,10 +19,10 @@ class WildcardQuery:
 
     @deprecated(
         version="0.0.12",
-        reason="`clp_ffi_py.wildcard_query.WildcardQuery` is supposed to be an abstract class and"
-        " should not be used directly. To create a wildcard query, please explicit instantiate"
-        " `clp_ffi_py.wildcard_query.SubstringWildcardQuery` or"
-        " `clp_ffi_py.wildcard_query.FullStringWildcardQuery`.",
+        reason=":class:`~clp_ffi_py.wildcard_query.WildcardQuery` is supposed to be an abstract"
```
kirkrodrigues: While we're on the topic, I think it's worth improving the docstrings in this file.

`WildcardQuery`:

```
    An abstract class defining a wildcard query. Users should instantiate a
    wildcard query through :class:`SubstringWildcardQuery` or
    :class:`FullStringWildcardQuery`.

    A wildcard string may contain the following types of wildcards:

    1. '*': match 0 or more characters.
    2. '?': match any single character.

    Each wildcard can be escaped using a preceding '\\\\' (a single backslash).
    Other characters that are escaped are treated as normal characters.
```

`WildcardQuery().__init__.deprecated`:

```
        reason=":class:`WildcardQuery` will soon be made abstract and should"
        " not be used directly. To create a wildcard query, instead use"
        " :class:`SubstringWildcardQuery` or :class:`FullStringWildcardQuery`."
```

`SubstringWildcardQuery`:

```
    A wildcard query that can match a substring in a log event's message, in
    contrast with :class:`FullStringWildcardQuery` where the needs to match the
    entire message.

    This class is derived from :class:`WildcardQuery` by adding both a prefix
    and a postfix wildcard ("*") to the input wildcard string.
```

`FullStringWildcardQuery`:

```
    A wildcard query where the query must match the entirety of the log event's
    message, in contrast with :class:`SubstringWildcardQuery` where the query
    only needs to match a substring.

    This class is derived from :class:`WildcardQuery` as a more explicit
    interface for full-string matches.

    Users can create a match that's anchored to only one end of the message by
    adding a prefix OR postfix wildcard ("*").
```

All `case_sensitive` params:

```
        :param case_sensitive: Whether to perform case-sensitive matching.
```

kirkrodrigues (changes requested):

clp_ffi_py/wildcard_query.py
LinZhihao-723: make sense. Pushed

kirkrodrigues (approved):
For the PR title, how about:

Docs: Clean-up docstrings for wildcard query methods and classes introduced in #62.

LinZhihao-723:
> For the PR title, how about:
> 
> Docs: Clean-up docstrings for wildcard query methods and classes introduced in #62.

lgtm

PR #67, opened 2024-07-07
Refactor source files to move towards latest coding standard:

LinZhihao-723:
# References

# Description

This PR refactors source files to meet our latest coding guidelines.
- For both C++ and Python files, we now use 100-char limit for comments.
- For C++, we drop `.inc` files and move the implementation into the header files.
- Update C++ header guards to make them consistent with CLP core.

# Validation performed

- Ensure all linters passed.
- Manually checked the generated doc site is properly rendered.


src/clp_ffi_py/ExceptionFFI.hpp
```diff
@@ -1,15 +1,14 @@
-#ifndef CLP_FFI_PY_EXCEPTION_FFI
-#define CLP_FFI_PY_EXCEPTION_FFI
+#ifndef CLP_FFI_PY_IR_EXCEPTION_FFI
+#define CLP_FFI_PY_IR_EXCEPTION_FFI
```
davidlion: This change seems incorrect no? (same with the endif)

LinZhihao-723: hmmm right, mb

davidlion (approved):
Refactor source files to move towards latest coding standard:
  - For C++ and Python files, use 100-char limit for comments.
  - For C++, drop .inc files and move the implementations into the header files.
  - Make C++ header guards consistent with CLP core.

PR #69, opened 2024-08-07
gh-actions: Change linting workflow schedule from daily to weekly.

LinZhihao-723:
# References

# Description

The current scheduled linting workflow runs too frequently and it frequently fails due to intensive workloads in GitHub. This failure could even propagate to the forked repos, which will bring confusion to the developers. As a solution, we decide to switch to weekly execution. Tuesday is chosen to avoid potential high GitHub traffic on Monday or Sunday.
Reference: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule

# Validation performed

Ensure the current workflow passes.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -30,7 +30,7 @@ on:
       - "src/**"
       - "tests/**"
   schedule:
-    # Run at midnight UTC every Tuesday with 15 minutes delay added to avoid high load periods
+    # Run every Tuesday at 00:15 UTC (the 15 is to avoid periods of high load)
     - cron: "15 0 * * 2"
 
 concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/build_wheels.yml
```diff
@@ -30,8 +30,8 @@ on:
       - "src/**"
       - "tests/**"
   schedule:
-    # Run at midnight UTC every day with 15 minutes delay added to avoid high load periods
-    - cron: "15 0 * * *"
+    # Run at midnight UTC every Tuesday with 15 minutes delay added to avoid high load periods
```
kirkrodrigues:
```suggestion
    # Run every Tuesday at 00:15 UTC (the 15 is to avoid periods of high load)
```

kirkrodrigues (dismissed):
Minor suggestion.

For the PR title, how about:

gh-actions: Change linting workflow schedule from daily to weekly.

kirkrodrigues (approved):

PR #73, opened 2024-10-23
Replace `setuptools` with `scikit-build-core` and CMake.

LinZhihao-723:
# References

# Description

This PR replaces `setuptools` by [scikit-build-core](https://scikit-build-core.readthedocs.io/en/latest/index.html). `scikit-build-core` uses CMake to build C-extensions and install the built libraries. This PR adds a CMake file following the guidelines from `scikit-build-core`'s official documentation. As a result, the old setuptool-based build files are all removed, including `setup.py` and `MANIFEST`. With the new build system, the manual build should still work (with `python3 -m build`), as well as `cibuildwheel` GitHub workflow. This PR also adds linting guideline + CI check for CMake configs.
Something to notice:
- After switching to `scikit-build-core`, we still relies on gitsubmodule to manage dependencies at this moment.
- The current way of packaging source distribution will include all files from open source CLP, making the source distribution larger than using `MANIFEST` with `setuptools`.
- The CMake file is supposed to work with `sci-kit-build-core` to build and install C-extension shared libraries properly. To enable us loading the CMake project into CLion easily, or to generate a compile DB, we also make it work without `scikit-build-core`, but in that way the libraries can't be automatically installed.

# Validation performed

- Ensure all existing workflow passed (this ensures the built wheels can be successfully installed, with all unit tests passed)
- Tested on the local dev environment to make sure:
  - The library can be installed successfully from a source distribution (indicating that the source distribution contains all the necessary files to build the library), with all unit tests passed
  - The library can be installed successfully from a locally built wheel file, with all unit tests passed
- Ensure the CMake project can be successfully loaded by CLion

## Summary by CodeRabbit

- **New Features**
  - Transitioned the build system from setuptools to scikit-build-core for better integration with CMake.
  - Introduced specific CMake configurations for enhanced project compatibility.
  - Added a new development dependency on `gersemi` for formatting CMake files.

- **Bug Fixes**
  - Improved compatibility with macOS by updating the build environment.

- **Chores**
  - Removed the `setup.py` file as part of the transition to the new build system.
  - Updated the GitHub Actions workflow to include a new linting step for CMake files.
  - Deleted the `MANIFEST.in` file, affecting package distribution inclusions.
  - Updated the README to clarify Python version compatibility and contributor guidelines.
  - Added a new environment variable and sources to the `Taskfile.yml` for improved build process.
  - Added a new entry to `.gitignore` for build artifacts generated by CMake.


CMakeLists.txt
```diff
@@ -0,0 +1,96 @@
...
+    set(CLP_FFI_PY_VERSION ${SKBUILD_PROJECT_VERSION} CACHE STRING
+            "Use scikit-build-core's project version." FORCE)
+else()
+    # The project is configured outside of scikit-build-core. Normally this is for build testing
+    # purpose, or loading the project into CLion. We don't install the built library in this config.
+    message(WARNING "Configured outside of scikit-build-core. Library installation is disabled.")
+    set(CLP_FFI_PY_INSTALL_LIBS OFF CACHE BOOL
+            "Disable library installation when building outside of scikit-build-core." FORCE)
+    set(CLP_FFI_PY "clp-ffi-py-local-build" CACHE STRING "Use local build name." FORCE)
+    set(CLP_FFI_PY_VERSION "0.0.12" CACHE STRING "Use local build version." FORCE)
```
junhaoliao: we probably don't need a specific version number here / don't want to update those in the future since we're specifying those fallbacks only for IDE setups?
```suggestion
    set(CLP_FFI_PY_VERSION "0.0.0" CACHE STRING "Use local build version." FORCE)
```

CMakeLists.txt
```diff
@@ -0,0 +1,96 @@
+# Minimum cmake version required by scikit-build-core: 3.15
```
junhaoliao: I feel we can remove this and only keep the 3.17 comment

CMakeLists.txt
```diff
@@ -0,0 +1,96 @@
+# Minimum cmake version required by scikit-build-core: 3.15
+# Minimum cmake version required by `Python_SOABI`: 3.17
+cmake_minimum_required(VERSION 3.17)
```
junhaoliao: If I remember correctly, the default CMake version on Ubuntu 20.04 is 3.16. Do we need to specify Ubuntu 22.04 LTS in the `Requirements` section in the README?

CMakeLists.txt
```diff
@@ -0,0 +1,96 @@
...
+else()
+    # The project is configured outside of scikit-build-core. Normally this is for build testing
+    # purpose, or loading the project into CLion. We don't install the built library in this config.
+    message(WARNING "Configured outside of scikit-build-core. Library installation is disabled.")
+    set(CLP_FFI_PY_INSTALL_LIBS OFF CACHE BOOL
+            "Disable library installation when building outside of scikit-build-core." FORCE)
+    set(CLP_FFI_PY "clp-ffi-py-local-build" CACHE STRING "Use local build name." FORCE)
+    set(CLP_FFI_PY_VERSION "0.0.12" CACHE STRING "Use local build version." FORCE)
+    # Enable compile commands by default if the generator supports it.
+    if (NOT CMAKE_EXPORT_COMPILE_COMMANDS AND CMAKE_GENERATOR MATCHES "Ninja|Unix Makefiles")
```
junhaoliao: I could be wrong but it's probably fine to enable CMAKE_EXPORT_COMPILE_COMMANDS even if the generator doesn't support it.

CMakeLists.txt
```diff
@@ -0,0 +1,96 @@
...
+    ${CLP_SRC_ROOT}/ffi/ir_stream/utils.cpp
+    ${CLP_SRC_ROOT}/ffi/encoding_methods.cpp
+    ${CLP_SRC_ROOT}/ir/parsing.cpp
+    ${CLP_SRC_ROOT}/ReaderInterface.cpp
+)
+
+target_compile_features(${CLP_FFI_PY_LIB_IR}
+    PRIVATE
+    cxx_std_20
+)
```
junhaoliao: Would it be better to place this before the source files / includes definitions? (probably right after `python_add_library`

CMakeLists.txt
```diff
@@ -0,0 +1,96 @@
...
+set(CMAKE_POSITION_INDEPENDENT_CODE ON CACHE BOOL "Compile libraries using PIC.")
+
+# Find Python interpreter, compiler and development environment.
+find_package(Python COMPONENTS Interpreter Development.Module REQUIRED)
+
+set(CPP_SRC_ROOT "src" CACHE STRING "Path to the root of all C++ source code")
+set(CLP_FFI_PY_SRC_ROOT "${CPP_SRC_ROOT}/clp_ffi_py" CACHE STRING
+        "Path to the root of clp-ffi-py's C++ source code")
+set(CLP_CORE_ROOT "src/clp/components/core" CACHE STRING "Path to CLP core's root")
+set(CLP_SRC_ROOT "${CLP_CORE_ROOT}/src/clp" CACHE STRING "Path to clp's source code")
```
junhaoliao: Those do not need to be `CACHE` variables, right? Fine to keep if we are using the syntax for variable docstrings.

CMakeLists.txt
```diff
@@ -0,0 +1,96 @@
...
+    LANGUAGES CXX C
+    VERSION ${CLP_FFI_PY_VERSION}
+)
+
+# C-extension modules will be compiled as shared libraries. To enable dynamic linking, all libraries
+# (including the static ones) must be compiled with position-independent-code option.
+set(CMAKE_POSITION_INDEPENDENT_CODE ON CACHE BOOL "Compile libraries using PIC.")
+
+# Find Python interpreter, compiler and development environment.
+find_package(Python COMPONENTS Interpreter Development.Module REQUIRED)
```
junhaoliao: Shall we place the `REQUIRED` keyword before specifying the components?

```suggestion
find_package(Python REQUIRED COMPONENTS Interpreter Development.Module)
```

CMakeLists.txt
```diff
@@ -0,0 +1,96 @@
...
+
+set(CPP_SRC_ROOT "src" CACHE STRING "Path to the root of all C++ source code")
```
junhaoliao: Would it be okay if we simply name it as `SRC_ROOT`?

pyproject.toml
```diff
@@ -1,10 +1,6 @@
 [build-system]
-requires = [
-    "setuptools >= 59.0",
-    "setuptools_scm",
-    "wheel",
-]
-build-backend = "setuptools.build_meta"
+requires = ["scikit-build-core"]
+build-backend = "scikit_build_core.build"
```
junhaoliao: nit: alphabetize (also good to place lists as the last items
```suggestion
build-backend = "scikit_build_core.build"
requires = ["scikit-build-core"]
```

pyproject.toml
```diff
@@ -68,6 +64,9 @@ archs = ["auto", "aarch64"]
 [tool.cibuildwheel.macos]
 archs = ["x86_64", "universal2", "arm64"]
 
+[tool.cibuildwheel.macos.environment]
+MACOSX_DEPLOYMENT_TARGET = "10.15"
```
junhaoliao, line 69: Is it because `std::filesystem::path` is only available after 10.15? Can we document this?

CMakeLists.txt
LinZhihao-723: https://scikit-build-core.readthedocs.io/en/latest/configuration.html#cmake-and-ninja-minimum-versions
If we set a minimum version here, `scikit-build-core` will download the compatible cmake via PyPI if possible

CMakeLists.txt
LinZhihao-723: I just realized we can set it to 3.15. If the version is lower than 3.26, `scikit-build-core` will ensure FindPython is backported: https://github.com/scikit-build/scikit-build-core

CMakeLists.txt
LinZhihao-723: This is copied from clp-ffi-go's CMake: https://github.com/y-scope/clp-ffi-go/blob/main/cpp/CMakeLists.txt#L8.
But you are right that the flag will be ignored if not supported: https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html

CMakeLists.txt
LinZhihao-723: I will refactor this part in the coming commits

CMakeLists.txt
LinZhihao-723: They don't have to be `CACHE` variables. I'm using this for docstrings. clp-ffi-go's cmake did the same thing: https://github.com/y-scope/clp-ffi-go/blob/main/cpp/CMakeLists.txt#L33

CMakeLists.txt
LinZhihao-723: Sure, make sense

pyproject.toml
LinZhihao-723: sure, make sense

pyproject.toml
LinZhihao-723: I moved this change to #74 
We need this env variable to enable >=C++17

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+        "Enable/Disable output of compile commands during generation."
+        FORCE
+    )
+endif()
+
+project(
+    ${CLP_FFI_PY}
+    LANGUAGES
+        CXX
+        C
```
junhaoliao: nit:
```suggestion
        C
        CXX
```

junhaoliao (dismissed):
LGTM. Feel free to have someone more knowledgeable about CMake to take a look before merging.

README.md
```diff
@@ -321,7 +318,6 @@ other package management tools such as `apt-get`:
 [6]: https://clang.llvm.org/docs/ClangFormatStyleOptions.html
 [7]: https://cibuildwheel.readthedocs.io/en/stable/
 [8]: https://clang.llvm.org/extra/clang-tidy/
-[9]: https://github.com/rizsotto/Bear
```
junhaoliao, line 329: Do we want to reshuffle the indices here?

LinZhihao-723: sigh, good point... I will reuse this index [9] for `gersimi` (which I forgot to add)

.github/workflows/build_wheels.yml
```diff
@@ -63,6 +63,9 @@ jobs:
       - run: |
           find src/clp_ffi_py/ -type f | xargs clang-format --dry-run --Werror
 
+      - run: |
+          gersemi -c -l 100 --list-expansion favour-expansion CMakeLists.txt
```
kirkrodrigues: Let's use the long form of the options:
- `--check`
- `--line-length`

README.md
```diff
@@ -294,6 +294,8 @@ formatting tools (found in [pyproject.toml]):
 * [ruff][10]: `ruff check --fix clp_ffi_py tests`
   * This performs linting according to PEPs. You should review and add any
     changes to your PR.
+* [gersemi][9]: `gersemi -i -l 100 --list-expansion favour-expansion CMakeLists.txt`
+  * This formats the cmake config. You should review and add any changes to your PR.
```
kirkrodrigues:
```suggestion
  * This formats CMakeLists.txt. You should review and add any changes to your PR.
```

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
+# Minimum cmake version required by scikit-build-core: 3.15
+cmake_minimum_required(VERSION 3.15)
```
kirkrodrigues: Why do we want to support such a low version of CMake? Folks usually [recommend](https://cliutils.gitlab.io/modern-cmake/chapters/basics.html#minimum-version) choosing a version of CMake that's at least as new as the compiler we support.

We should also specify the maximum version that we've tested with using the range synta `A..B`.

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
+# Minimum cmake version required by scikit-build-core: 3.15
+cmake_minimum_required(VERSION 3.15)
+
+set(CLP_FFI_PY_INSTALL_LIBS ON CACHE BOOL "Whether to install the built libraries." FORCE)
```
kirkrodrigues: Can we move this into the `if` as:
`set(CLP_FFI_PY_INSTALL_LIBS ON CACHE BOOL "Enable installing the built libraries." FORCE)`

It's a bit awkward to set this outside and then override it only in the `else`.

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
+# Minimum cmake version required by scikit-build-core: 3.15
+cmake_minimum_required(VERSION 3.15)
+
+set(CLP_FFI_PY_INSTALL_LIBS ON CACHE BOOL "Whether to install the built libraries." FORCE)
+
+if(DEFINED SKBUILD_PROJECT_NAME)
+    set(CLP_FFI_PY
```
kirkrodrigues: How about `CLP_FFI_PY_PROJECT_NAME`?

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
+# Minimum cmake version required by scikit-build-core: 3.15
+cmake_minimum_required(VERSION 3.15)
+
+set(CLP_FFI_PY_INSTALL_LIBS ON CACHE BOOL "Whether to install the built libraries." FORCE)
+
+if(DEFINED SKBUILD_PROJECT_NAME)
+    set(CLP_FFI_PY
+        ${SKBUILD_PROJECT_NAME}
+        CACHE STRING
+        "Use scikit-build-core's project name."
```
kirkrodrigues:
```suggestion
        "The name of the project parsed by scikit-build-core."
```

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+    set(CLP_FFI_PY
+        ${SKBUILD_PROJECT_NAME}
+        CACHE STRING
+        "Use scikit-build-core's project name."
+        FORCE
+    )
+    set(CLP_FFI_PY_VERSION
+        ${SKBUILD_PROJECT_VERSION}
+        CACHE STRING
+        "Use scikit-build-core's project version."
```
kirkrodrigues:
```suggestion
        "The version of the project parsed by scikit-build-core."
```

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
+# Minimum cmake version required by scikit-build-core: 3.15
+cmake_minimum_required(VERSION 3.15)
+
+set(CLP_FFI_PY_INSTALL_LIBS ON CACHE BOOL "Whether to install the built libraries." FORCE)
+
+if(DEFINED SKBUILD_PROJECT_NAME)
```
kirkrodrigues:
```suggestion
# Define the variables necessary to build and install the project differently depending on whether
# the project is being configured from scikit-build-core. If configured from outside
# scikit-build-core, we need to use placeholder values for the variables that are usually defined
# by scikit-build-core.
if(DEFINED SKBUILD_PROJECT_NAME)
```

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+    set(CLP_FFI_PY_VERSION
+        ${SKBUILD_PROJECT_VERSION}
+        CACHE STRING
+        "Use scikit-build-core's project version."
+        FORCE
+    )
+else()
+    # The project is configured outside of scikit-build-core. Normally this is for build testing
+    # purpose, or loading the project into CLion. We don't install the built library in this config.
+    message(WARNING "Configured outside of scikit-build-core. Library installation is disabled.")
```
kirkrodrigues: I think we can drop this comment in favour of the one I suggested outside the if-else. I think we can also drop this log since even if you wanted to install the library through your IDE, you still wouldn't be able to generate a wheel (which is what scikit-build seems to do), so I don't really see why you would want to install the library in the first place.

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+    set(CLP_FFI_PY_INSTALL_LIBS
+        OFF
+        CACHE BOOL
+        "Disable library installation when building outside of scikit-build-core."
```
kirkrodrigues:
```suggestion
        "Disable installing the built library."
```

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+        FORCE
+    )
+    set(CLP_FFI_PY "clp-ffi-py-local-build" CACHE STRING "Use local build name." FORCE)
```
kirkrodrigues:
```suggestion
    set(CLP_FFI_PY "clp-ffi-py" CACHE STRING "Use a placeholder project name." FORCE)
```

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+    set(CLP_FFI_PY_VERSION "0.0.0" CACHE STRING "Version info is unused for local build." FORCE)
```
kirkrodrigues:
```suggestion
    set(CLP_FFI_PY_VERSION "0.0.0" CACHE STRING "Use a placeholder version." FORCE)
```

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+        CACHE BOOL
+        "Enable/Disable output of compile commands during generation."
+        FORCE
+    )
+endif()
+
+project(
+    ${CLP_FFI_PY}
+    LANGUAGES
+        C
```
kirkrodrigues:
```suggestion
```

Unless we have C source files in the project?

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+        CXX
+    VERSION ${CLP_FFI_PY_VERSION}
+)
+
+# C-extension modules will be compiled as shared libraries. To enable dynamic linking, all libraries
+# (including the static ones) must be compiled with position-independent-code option.
```
kirkrodrigues:
```suggestion
# (including the static ones) must be compiled with the position-independent-code option.
```

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+set(CMAKE_POSITION_INDEPENDENT_CODE ON CACHE BOOL "Compile libraries using PIC.")
+
+# Find Python interpreter, compiler and development environment.
```
kirkrodrigues:
```suggestion
```

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+find_package(
+    Python
+    REQUIRED
+    COMPONENTS
+        Interpreter
+        Development.Module
+)
+
+set(CLP_FFI_PY_LIB_IR "native" CACHE STRING "The name of IR native library")
```
kirkrodrigues: How about `CLP_FFI_PY_LIB_NAME`?

kirkrodrigues:
```suggestion
set(CLP_FFI_PY_LIB_IR "native")
```

I don't think we need to use a docstring for this, do we? I guess maybe it aids debugging if you wanted to print out all the variables used in the script, but documenting every variable just seems like an excessive amount of work. If I needed to print out all variables, I would rather define a simple CMake function to do so.

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+python_add_library(
+    ${CLP_FFI_PY_LIB_IR}
+    MODULE
+    WITH_SOABI
+)
+
+target_compile_features(${CLP_FFI_PY_LIB_IR} PRIVATE cxx_std_20)
+
+set(SRC_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/src" CACHE STRING "Root path of C++ source code")
```
kirkrodrigues: How about `CLP_FFI_PY_SRC_DIR`?

We've been trying to prefix all variables defined by us with the project name. This avoids conflicts with variables defined by things we include/use/etc.

kirkrodrigues:
```suggestion
set(SRC_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/src")
```

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+set(CLP_FFI_PY_SRC_ROOT
```
kirkrodrigues: How about `CLP_FFI_PY_LIB_SRC_DIR`?

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+    "${SRC_ROOT}/clp_ffi_py"
+    CACHE STRING
+    "Root path of clp-ffi-py's C++ source code"
```
kirkrodrigues:
```suggestion
```

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+)
+set(CLP_CORE_ROOT "${SRC_ROOT}/clp/components/core" CACHE STRING "Root path of CLP core")
```
kirkrodrigues: How about `CLP_FFI_PY_CLP_CORE_SRC_DIR`?

kirkrodrigues:
```suggestion
set(CLP_CORE_ROOT "${SRC_ROOT}/clp/components/core")
```

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+
+add_subdirectory(${CLP_CORE_ROOT}/src/clp/string_utils)
+
+# Source files of clp-ffi-core
+# Note: we don't add headers hear since they are not considered to be a part of the project.
```
kirkrodrigues: Why aren't they considered part of the project? I know CMake ignores header files, so I would imagine the only reason to add header files is if your IDE needs them to understand the project.

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+set(CLP_FFI_PY_SRC_ROOT
+    "${SRC_ROOT}/clp_ffi_py"
+    CACHE STRING
+    "Root path of clp-ffi-py's C++ source code"
+)
+set(CLP_CORE_ROOT "${SRC_ROOT}/clp/components/core" CACHE STRING "Root path of CLP core")
+
+add_subdirectory(${CLP_CORE_ROOT}/src/clp/string_utils)
+
+# Source files of clp-ffi-core
```
kirkrodrigues:
```suggestion
```

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+# Note: we don't add headers hear since they are not considered to be a part of the project.
+set(SOURCE_FILES_CLP_FFI_CORE
```
kirkrodrigues: How about `CLP_FFI_PY_CLP_CORE_SOURCES`?

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+    ${CLP_CORE_ROOT}/src/clp/BufferReader.cpp
+    ${CLP_CORE_ROOT}/src/clp/ffi/ir_stream/decoding_methods.cpp
+    ${CLP_CORE_ROOT}/src/clp/ffi/ir_stream/encoding_methods.cpp
+    ${CLP_CORE_ROOT}/src/clp/ffi/ir_stream/utils.cpp
+    ${CLP_CORE_ROOT}/src/clp/ffi/encoding_methods.cpp
+    ${CLP_CORE_ROOT}/src/clp/ir/parsing.cpp
+    ${CLP_CORE_ROOT}/src/clp/ReaderInterface.cpp
+)
+
+# Header files and source files of clp-ffi-py
```
kirkrodrigues:
```suggestion
```

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+# Note: headers should be included to allow CLion loads the project properly.
+set(SOURCE_FILES_CLP_FFI_PY_LIB_IR
```
kirkrodrigues: How about `CLP_FFI_PY_LIB_SOURCES`?

kirkrodrigues: For me, I don't need to specify the header files in order for CLion to pick them up. That said, if I open a header file directly, CLion may not understand that it's part of the project until I open a file where it's included. I'm fine if we want to list headers in the CMakeLists.txt but then we should probably do it for the CLP sources too, right?

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+)
+
+target_include_directories(
+    ${CLP_FFI_PY_LIB_IR}
+    SYSTEM
+    PRIVATE
+        ${CLP_CORE_ROOT}/src
+        ${CLP_CORE_ROOT}/submodules
+        ${SRC_ROOT}
+)
```
kirkrodrigues:
```suggestion
# NOTE: We mark the include directories below as system headers so that the compiler (including
# `clang-tidy`) doesn't generate warnings from them.
target_include_directories(
    ${CLP_FFI_PY_LIB_IR}
    SYSTEM
    PRIVATE
        ${CLP_CORE_ROOT}/src
        ${CLP_CORE_ROOT}/submodules
)

target_include_directories(
    ${CLP_FFI_PY_LIB_IR}
    PRIVATE
        ${SRC_ROOT}
)
```

CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+target_link_libraries(${CLP_FFI_PY_LIB_IR} PRIVATE clp::string_utils)
+
+if(CLP_FFI_PY_INSTALL_LIBS)
+    install(
+        TARGETS
+            ${CLP_FFI_PY_LIB_IR}
+            DESTINATION
+            ${CLP_FFI_PY}/ir
+            COMPONENT
+            PythonModule
```
kirkrodrigues: What does this do / where is this documented?

pyproject.toml
```diff
@@ -97,3 +92,15 @@ line-length = 100
 [tool.ruff.lint]
 select = ["E", "I", "F"]
 isort.order-by-type = false
+
+[tool.scikit-build]
+cmake.build-type = "Release"
+cmake.version = ">=3.15"
```
kirkrodrigues: The docs say:

> The default on 0.10+ is "CMakeLists.txt", which will read it from the project's CMakeLists.txt file, or ">=3.15" if unreadable or <0.10.

So I think it's safer not to specify it?

pyproject.toml
```diff
@@ -97,3 +92,15 @@ line-length = 100
...
+# By default, sdist-build-core uses `.gitignore` to select what to exclude from the source
+# distribution. This list is to explict exclude files not specified in `.gitignore`.
```
kirkrodrigues:
```suggestion
# distribution. This list is to explictly exclude files not specified in `.gitignore`.
```

pyproject.toml
```diff
@@ -97,3 +92,15 @@ line-length = 100
 [tool.ruff.lint]
 select = ["E", "I", "F"]
 isort.order-by-type = false
+
+[tool.scikit-build]
```
kirkrodrigues, line 97: Might be useful to enable `logging.level = "INFO"`.

CMakeLists.txt
kirkrodrigues, line 1: Let's also modify the Taskfile:
- Add `CMakeLists.txt` and `pyproject.toml` to the sources of `wheel`.
- Add `export SKBUILD_BUILD_DIR="{{.WHEEL_BUILD_DIR}}"` before `python3 -m build`.

pyproject.toml
```diff
@@ -69,7 +65,6 @@ archs = ["auto", "aarch64"]
 archs = ["x86_64", "universal2", "arm64"]
 
 [tool.cibuildwheel.macos.environment]
-# This should be kept in sync with the environment variable in setup.py.
 MACOSX_DEPLOYMENT_TARGET = "10.15"
```
kirkrodrigues, line 69:
```suggestion
# We need 10.15 since CLP core uses std::from_chars.
MACOSX_DEPLOYMENT_TARGET = "10.15"
```

pyproject.toml
```diff
@@ -97,3 +92,15 @@ line-length = 100
 [tool.ruff.lint]
 select = ["E", "I", "F"]
 isort.order-by-type = false
+
+[tool.scikit-build]
+cmake.build-type = "Release"
+cmake.version = ">=3.15"
+# By default, sdist-build-core uses `.gitignore` to select what to exclude from the source
+# distribution. This list is to explict exclude files not specified in `.gitignore`.
+sdist.exclude = [
```
kirkrodrigues, line 102: Do you know why the source distribution doesn't include the non-source files in clp? It's not a problem, but I just don't understand how it knows.

kirkrodrigues (changes requested):

CMakeLists.txt
kirkrodrigues, line 1: Also, this file is using CRLF instead of LF.

CMakeLists.txt
LinZhihao-723: The reason to include the source files from clp core for compilation is we don't have clp-ffi-core library yet. Strictly speaking clp-ffi-core is an external library that we use; we're not supposed to modify any source/header files. As a result, the headers shouldn't be a part of the project. This doesn't affect CLion's code navigation, i.e., you can still jump to the declaration of a clp-ffi-core function in the IDE because CLion resolves the include path. But these headers are just like any C++ standard library headers, you're not supposed to make any changes, and CLion will try to tell you they are not a part of the project.

CMakeLists.txt
LinZhihao-723: 1. I'd prefer to add headers here
2. CLP sources are different, as explained in the reply of the previous comment

CMakeLists.txt
LinZhihao-723: GCC 10.1 2020 
clang 7.0 2018
MSVC 1930 2022

I will set the minimum requirement to 3.26, which is formally released in 2023. It should be available through pip and scikit-build-core will resolve it if the system's version is lower than 3.26.

pyproject.toml
LinZhihao-723: true, make sense

CMakeLists.txt
LinZhihao-723: Sorry it's from this example and I directly copied the install command: https://github.com/scikit-build/scikit-build-core/blob/main/tests/packages/simplest_c/CMakeLists.txt
Seems like it might not be necessary

CMakeLists.txt
LinZhihao-723: Prefered to use `CLP_FFI_PY_LIB_IR_SOURCES` in case we have other C-extension libraries in the future

CMakeLists.txt
LinZhihao-723: Fixed using `sed -i 's/\r$//' CMakeLists.txt`, should be fixed unless god gpt lies to me again, lol

LinZhihao-723: Done. Also realized I need to update the workflow file to remove "setup.py" and include "CMakeLists.txt"

pyproject.toml
LinZhihao-723: I think it's because this exclude list applies to very repo (including submodules). I tried to change `cmake-build-*` to `cmake*`, then the `cmake` directory under clp/components/core would also be excluded.

CMakeLists.txt
kirkrodrigues: What's the minimum compiler version we're supporting?

CMakeLists.txt
```diff
@@ -0,0 +1,136 @@
...
+        FORCE
+    )
+    set(CLP_FFI_PY_VERSION
+        ${SKBUILD_PROJECT_VERSION}
+        CACHE STRING
+        "The version of the project parsed by scikit-build-core."
+        FORCE
+    )
+else()
+    set(CLP_FFI_PY_INSTALL_LIBS OFF CACHE BOOL "Disable installing the built library." FORCE)
```
kirkrodrigues:
```suggestion
    set(CLP_FFI_PY_INSTALL_LIBS OFF CACHE BOOL "Disable installing the built libraries." FORCE)
```

CMakeLists.txt
kirkrodrigues: @LinZhihao-723 Seems this comment wasn't addressed?

CMakeLists.txt
```diff
@@ -0,0 +1,136 @@
...
+
+target_compile_features(${CLP_FFI_PY_LIB_IR} PRIVATE cxx_std_20)
+
+set(SRC_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/src")
+set(CLP_FFI_PY_LIB_SRC_DIR "${SRC_ROOT}/clp_ffi_py")
+set(CLP_CORE_ROOT "${SRC_ROOT}/clp/components/core")
+
+add_subdirectory(${CLP_CORE_ROOT}/src/clp/string_utils)
+
+# Note: we don't add headers here since they are not considered to be a part of the project.
```
kirkrodrigues:
```suggestion
# NOTE: We don't add headers here since CLP core is technically a library we're using, not a part of
# this project.
```

CMakeLists.txt
```diff
@@ -0,0 +1,136 @@
...
+    ${CLP_CORE_ROOT}/src/clp/BufferReader.cpp
+    ${CLP_CORE_ROOT}/src/clp/ffi/ir_stream/decoding_methods.cpp
+    ${CLP_CORE_ROOT}/src/clp/ffi/ir_stream/encoding_methods.cpp
+    ${CLP_CORE_ROOT}/src/clp/ffi/ir_stream/utils.cpp
+    ${CLP_CORE_ROOT}/src/clp/ffi/encoding_methods.cpp
+    ${CLP_CORE_ROOT}/src/clp/ir/parsing.cpp
+    ${CLP_CORE_ROOT}/src/clp/ReaderInterface.cpp
+)
+
+# Note: headers should be included to allow CLion loads the project properly.
```
kirkrodrigues:
```suggestion
# NOTE: We include headers to ensure IDEs like CLion load the project properly.
```

pyproject.toml
```diff
@@ -97,3 +93,15 @@ line-length = 100
 [tool.ruff.lint]
 select = ["E", "I", "F"]
 isort.order-by-type = false
+
+[tool.scikit-build]
+cmake.build-type = "Release"
+logging.level = "INFO"
+# By default, sdist-build-core uses `.gitignore` to select what to exclude from the source
+# distribution. This list is to explictly exclude files not specified in `.gitignore`.
```
kirkrodrigues:
```suggestion
# distribution. This list is to explicitly exclude files not specified in `.gitignore`.
```

kirkrodrigues (changes requested):

pyproject.toml
```diff
@@ -97,3 +93,15 @@ line-length = 100
...
 isort.order-by-type = false
+
+[tool.scikit-build]
+cmake.build-type = "Release"
+logging.level = "INFO"
+# By default, sdist-build-core uses `.gitignore` to select what to exclude from the source
+# distribution. This list is to explicitly exclude files not specified in `.gitignore`.
+sdist.exclude = [
+    ".*",
+    "cmake-build-*",
```
kirkrodrigues: Why not add this to gitignore instead? I don't think we have anything we want to commit that has this name?

CMakeLists.txt
```diff
@@ -0,0 +1,137 @@
...
+    ${CLP_FFI_PY_LIB_IR}
+    MODULE
+    WITH_SOABI
+)
+
+target_compile_features(${CLP_FFI_PY_LIB_IR} PRIVATE cxx_std_20)
+
+set(CLP_FFI_PY_SRC_DIR "${CMAKE_CURRENT_SOURCE_DIR}/src")
+set(CLP_FFI_PY_LIB_SRC_DIR "${CLP_FFI_PY_SRC_DIR}/clp_ffi_py")
+set(CLP_CORE_ROOT "${CLP_FFI_PY_SRC_DIR}/clp/components/core")
```
kirkrodrigues: As discussed offline and in the comment from the previous review, this should probably be `CLP_FFI_PY_CLP_CORE_SRC_DIR`.

kirkrodrigues (changes requested):

CMakeLists.txt
LinZhihao-723: Renamed to `CLP_FFI_PY_CLP_CORE_DIR`. Strictly speaking the source directory is under core directory, so it's not `SRC_DIR`

kirkrodrigues (approved):
For the PR title, how about:

```
Replace `setuptools` with `scikit-build-core` and CMake.
```

PR #76, opened 2024-10-25
feat(linting): Add Taskfile tasks to run linters.

haiqi96:
# References

# Description
Port latest task-based linter flow into clp ffi py. Made a few customizations:
1. Support passing target directory as a list to the sub-jobs
2. Add mypy and docformatter to python formatting tasks.
3. Temporarily disabled cpp-static-check because the CPP source files still contains warnings which get treated as errors.
4. Updated the git flow to use the new linter tool

# Validation performed
Confirmed that linter runs locally.
Confirmed that git workflow's linter job still passes,

## Summary by CodeRabbit

- **New Features**
	- Introduced new submodule `tools/yscope-dev-utils` for enhanced utility functions.
	- Added a new configuration file `.gersemirc` for YAML linting with specific settings.

- **Bug Fixes**
	- Updated bug report template for improved clarity and structure.

- **Documentation**
	- Enhanced `README.md` with updated links, compatibility notes, and new sections on linting and file management.
	- Added a custom CSS file to the Sphinx documentation build process.

- **Chores**
	- Updated development dependencies in `requirements-dev.txt` to include `clang-tidy` and `yamllint`.
	- Modified GitHub Actions workflow to streamline linting processes.
	- Updated task definitions in `Taskfile.yml` for better organization and clarity.


.yamllint.yml
```diff
@@ -0,0 +1,51 @@
+extends: "default"
```
haiqi96: straight copy & paste from OSS

tools/yscope-dev-utils
haiqi96, line 1: straight copy & paste from OSS

tools/scripts/deps-download/init.sh
haiqi96, line 1: straight copy & paste from OSS

tools/scripts/deps-download/download-dep.py
haiqi96, line 1: straight copy & paste from OSS

lint-requirements.txt
haiqi96, line 1: straight copy & paste from OSS

kirkrodrigues:
See https://github.com/y-scope/spider for more up-to-date ways of doing things.

Taskfile.yml
```diff
@@ -1,23 +1,45 @@
...
+  docs: "docs/tasks.yml"
+  lint: "lint-tasks.yml"
+  utils: "tools/yscope-dev-utils/taskfiles/utils.yml"
 
 vars:
   # Paths
   BUILD_DIR: "{{.ROOT_DIR}}/build"
+  CLP_FFI_PY_BUILD_DIR: "{{.BUILD_DIR}}/clp_ffi_py"
+  CLP_FFI_PY_CMAKE_CACHE: "{{.CLP_FFI_PY_BUILD_DIR}}/CMakeCache.txt"
+  CLP_FFI_PY_COMPILE_COMMANDS_DB: "{{.CLP_FFI_PY_BUILD_DIR}}/compile_commands.json"
```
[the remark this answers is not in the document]
haiqi96: The G_BUILD_DIR has been renamed to BUILD_DIR

lint-tasks.yml
```diff
@@ -0,0 +1,219 @@
...
+    sources: &cpp_format_src_files
+      - "{{.LINT_VENV_CHECKSUM_FILE}}"
+      - "{{.CLP_FFI_PY_SRC_DIR}}/**/*.cpp"
+      - "{{.CLP_FFI_PY_SRC_DIR}}/**/*.h"
+      - "{{.CLP_FFI_PY_SRC_DIR}}/**/*.hpp"
+      - "{{.CLP_FFI_PY_SRC_DIR}}/**/*.inc"
+      - "src/.clang-format"
+      - "{{.TASKFILE}}"
+      - ".clang-format"
+    deps: ["cpp-configs", "venv"]
```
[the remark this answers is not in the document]
haiqi96: no they are not the same clang-format file

README.md
```diff
@@ -271,68 +277,80 @@ python -m unittest -bv
...
-other package management tools such as `apt-get`:
-
-* [clang-tidy][8]: `clang-tidy --extra-arg=-std=c++17 PATH_TO_THE_FILE`
-  * This static analysis tool catches improper coding behaviors based on the
-    rules specified in `.clang-tidy`, and sends suggestions corresponding to
-    each warning. Developers should manually review all the warnings and try
-    with their best effort to fix the reasonable ones.
+
+Note that `docformatter` should be run before `linter` to give Black the
+[last][7].
```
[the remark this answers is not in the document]
haiqi96: @LinZhihao-723 any preference on the 2nd comment?

kirkrodrigues:
@haiqi96 You're missing a PR description, right?

kirkrodrigues:
Also a good time to work on a PR title since you're waiting for review.

.github/ISSUE_TEMPLATE/feature-request.yml
haiqi96, line 1: Only indentation changes

.github/workflows/build_wheels.yml
```diff
@@ -38,198 +34,189 @@ concurrency:
...
-
-      - run: |
-          gersemi --check --line-length 100 --list-expansion favour-expansion CMakeLists.txt
+      - run: "task lint:check"
 
   build_wheels:
-    needs: [linters]
-    name: Build ${{ matrix.build.name }}
-    runs-on: ${{ matrix.build.os }}
+    needs: ["linters"]
```
haiqi96, line 55: Only indentation and quote changes below this line

.github/ISSUE_TEMPLATE/bug-report.yml
haiqi96, line 1: Only indentation and quote changes

.gersemirc
haiqi96, line 1: Straight copy from https://github.com/y-scope/spider

lint-tasks.yml
```diff
@@ -0,0 +1,219 @@
...
+        yamllint \
+          --config-file "tools/yscope-dev-utils/lint-configs/.yamllint.yml" \
+          --strict \
+          .gersemirc \
+          .github \
+          docs \
+          lint-tasks.yml \
+          src/.clang-format \
+          Taskfile.yml
+
```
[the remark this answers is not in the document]
haiqi96: I don't think we should use this cuz it will also target all yml files under the submodules

src/.clang-format
LinZhihao-723, line 1: Can we move this inside `src/clp_ffi_py`? My concern is the src directory also contains third party dependencies, including clp, GSL, etc. This config should only be applied to `clp_ffi_py`. On the other hand, we should send another PR to remove all dependencies into a standalone directory

requirements-dev.txt
LinZhihao-723, line 1: Can we have a separate file for linter dependencies? like `requirements-lint.txt`

docs/conf/conf.py
```diff
@@ -66,5 +66,6 @@
 # -- Theme custom CSS and JS ---------------------------------------------------
 # https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/static_assets.html
 
+
```
LinZhihao-723: Added by the linter?

Taskfile.yml
```diff
@@ -1,23 +1,47 @@
...
-  docs:
-    taskfile: "docs/tasks.yml"
+  docs: "docs/tasks.yml"
+  lint: "lint-tasks.yml"
+  utils: "tools/yscope-dev-utils/taskfiles/utils.yml"
 
 vars:
   # Paths
   BUILD_DIR: "{{.ROOT_DIR}}/build"
+  CLP_FFI_PY_BUILD_DIR: "{{.BUILD_DIR}}/clp_ffi_py"
```
LinZhihao-723: Using `clp_ffi_py` might not be a good idea. In this project, `clp_ffi_py` usually refers to the actual Python package, including the Python source and the installed ffi shared library. How about:
```suggestion
  CLP_FFI_PY_CMAKE_BUILD_DIR: "{{.BUILD_DIR}}/cmake_build"
```

Taskfile.yml
LinZhihao-723, line 1: This might be out of scope:
In `submodule` task, we have a checksum validation that only checks against CLP. Shall we fix that?

Taskfile.yml
```diff
@@ -1,23 +1,47 @@
 version: "3"
 
 includes:
-  docs:
-    taskfile: "docs/tasks.yml"
+  docs: "docs/tasks.yml"
+  lint: "lint-tasks.yml"
+  utils: "tools/yscope-dev-utils/taskfiles/utils.yml"
 
 vars:
```
LinZhihao-723, line 8: Seems like we are using `G_` prefix for global variables in the latest repos (https://github.com/y-scope/spider/blob/main/lint-tasks.yaml)
Shall we update to the new naming convention in this PR? @kirkrodrigues
If not this file, maybe at least the new task file `lint-tasks.yml`?

Taskfile.yml
```diff
@@ -1,23 +1,47 @@
...
   # Paths
   BUILD_DIR: "{{.ROOT_DIR}}/build"
+  CLP_FFI_PY_BUILD_DIR: "{{.BUILD_DIR}}/clp_ffi_py"
+  CLP_FFI_PY_CMAKE_CACHE: "{{.CLP_FFI_PY_BUILD_DIR}}/CMakeCache.txt"
+  CLP_FFI_PY_COMPILE_COMMANDS_DB: "{{.CLP_FFI_PY_BUILD_DIR}}/compile_commands.json"
+  CLP_FFI_PY_SRC_DIR: "{{.ROOT_DIR}}/src/clp_ffi_py"
```
LinZhihao-723: This is only the source directory for cpp file. How about:
`CLP_FFI_PY_CPP_SRC_DIR`?

lint-tasks.yml
```diff
@@ -0,0 +1,217 @@
...
+  cmake-fix:
+    deps: ["venv"]
+    cmds:
+      - task: "cmake"
+        vars:
+          FLAGS: "--in-place"
+
+  cpp-configs:
+    # TODO: remove this once all clang-tidy has been fixed
```
LinZhihao-723, line 40: Is this TODO added by mistake?

lint-tasks.yml
```diff
@@ -0,0 +1,217 @@
...
+  cpp-format-check:
+    sources: &cpp_format_src_files
+      - "{{.LINT_VENV_CHECKSUM_FILE}}"
+      - "{{.CLP_FFI_PY_SRC_DIR}}/**/*.cpp"
+      - "{{.CLP_FFI_PY_SRC_DIR}}/**/*.h"
+      - "{{.CLP_FFI_PY_SRC_DIR}}/**/*.hpp"
+      - "{{.CLP_FFI_PY_SRC_DIR}}/**/*.inc"
+      - "src/.clang-format"
+      - "{{.TASKFILE}}"
+      - ".clang-format"
```
LinZhihao-723, line 65: - Do we also need `tools/yscope-dev-utils/lint-configs/.clang-format`?
- Do we need `.clang-format` as it's a symlink?

lint-tasks.yml
```diff
@@ -0,0 +1,217 @@
...
+      # TODO: re-enable this once all clang-tidy has been fixed
+      # - task: "cpp-static-fix"
+
+  cpp-format-check:
+    sources: &cpp_format_src_files
+      - "{{.LINT_VENV_CHECKSUM_FILE}}"
+      - "{{.CLP_FFI_PY_SRC_DIR}}/**/*.cpp"
+      - "{{.CLP_FFI_PY_SRC_DIR}}/**/*.h"
+      - "{{.CLP_FFI_PY_SRC_DIR}}/**/*.hpp"
+      - "{{.CLP_FFI_PY_SRC_DIR}}/**/*.inc"
```
LinZhihao-723: I think we can safely drop `.inc` as this project doesn't have any inc file, and we're not planning to add more according to our latest guideline

lint-tasks.yml
```diff
@@ -0,0 +1,217 @@
...
+      - "{{.CLP_FFI_PY_CMAKE_CACHE}}"
+      - "{{.CLP_FFI_PY_COMPILE_COMMANDS_DB}}"
+      - "{{.TASKFILE}}"
+      - "Taskfile.yml"
+      - ".clang-tidy"
+    deps: [":config-cmake-project", "cpp-configs", "venv"]
+    cmds:
+      - task: ":utils:clang-tidy"
+        vars:
+          FLAGS: "--config-file=.clang-tidy -p {{.CLP_FFI_PY_COMPILE_COMMANDS_DB}}"
```
LinZhihao-723: Shall we explicitly use the config file under the root directory?

lint-tasks.yml
```diff
@@ -0,0 +1,217 @@
...
+    deps: ["venv"]
+    cmds:
+      - |-
+        . "{{.LINT_VENV_DIR}}/bin/activate"
+        yamllint \
+          --config-file "tools/yscope-dev-utils/lint-configs/.yamllint.yml" \
+          --strict \
+          .gersemirc \
+          .github \
+          docs \
```
LinZhihao-723: Maybe only `docs/tasks.yml`?

lint-tasks.yml
```diff
@@ -0,0 +1,217 @@
...
+      vars: ["SRC_DIRS"]
+    deps: ["venv"]
+    cmds:
+      - for:
+          var: "SRC_DIRS"
+        cmd: |-
+          . "{{.LINT_VENV_DIR}}/bin/activate"
+          mypy {{.ITEM}}
+
+  py:
```
LinZhihao-723, line 157: Can we use `dir` to enforce the task to run under the root directory, so that the linter configures specified in `pyproject.toml` can be automatically applied? In that case we don't need to specify black's line length explicitly

lint-tasks.yml
```diff
@@ -0,0 +1,217 @@
...
+
+  cmake:
+    internal: true
+    requires:
+      vars: ["FLAGS"]
+    cmd: |-
+      . "{{.LINT_VENV_DIR}}/bin/activate"
+      gersemi {{.FLAGS}} CMakeLists.txt {{.CLP_FFI_PY_SRC_DIR}}
+
+  mypy:
```
LinZhihao-723: Same for running under `dir` to use the config in `pyproject.toml`

lint-tasks.yml
```diff
@@ -0,0 +1,217 @@
...
+    cmds:
+      - task: "mypy"
+        vars:
+          SRC_DIRS:
+            ref: ".PYTHON_LINT_DIRS"
+      - task: "py"
+        vars:
+          DOCFORMATTER_FLAGS: "--check --diff"
+          BLACK_FLAGS: "--check"
+          RUFF_FLAGS: ""
```
LinZhihao-723:
```suggestion
          RUFF_FLAGS: "--output-format=github"
```
How about adding this to be consistent with our current workflow check?

lint-tasks.yml
```diff
@@ -0,0 +1,217 @@
...
+      - task: ":utils:clang-tidy"
+        vars:
+          FLAGS: "--config-file=.clang-tidy -p {{.CLP_FFI_PY_COMPILE_COMMANDS_DB}}"
+          SRC_PATHS:
+            ref: ".CPP_LINT_DIRS"
+          VENV_DIR: "{{.LINT_VENV_DIR}}"
+
+  py-check:
+    cmds:
+      - task: "mypy"
```
LinZhihao-723: Can we also add `mypy` check in `py-fix`?
- Normally when running `py-fix` people would also want to see mypy warnings
- We had same config for yml-check and yml-fix

If we decide to apply this comment, we should also be able to merge `mypy` task into py tasks

lint-tasks.yml
```diff
@@ -0,0 +1,217 @@
...
+        vars:
+          SRC_DIRS:
+            ref: ".PYTHON_LINT_DIRS"
+      - task: "py"
+        vars:
+          DOCFORMATTER_FLAGS: "--check --diff"
+          BLACK_FLAGS: "--check"
```
LinZhihao-723:
```suggestion
          BLACK_FLAGS: "--check --diff"
```
To be consistent with our current workflow check

README.md
```diff
@@ -271,69 +277,65 @@ python -m unittest -bv
...
+| `lint:cmake-check`      | Runs the CMake linters.                                  |
+| `lint:cmake-fix`        | Runs the CMake linters and fixes any violations.         |
+| `lint:cpp-check`        | Runs the C++ linters (formatters and static analyzers).  |
+| `lint:cpp-fix`          | Runs the C++ linters and fixes some violations.          |
+| `lint:cpp-format-check` | Runs the C++ formatters.                                 |
+| `lint:cpp-format-fix`   | Runs the C++ formatters and fixes some violations.       |
+| `lint:cpp-static-check` | Runs the C++ static analyzers.                           |
+| `lint:cpp-static-fix`   | Runs the C++ static analyzers and fixes some violations. |
+| `lint:yml-check`        | Runs the YAML linters.                                   |
+| `lint:yml-fix`          | Runs the YAML linters and fixes some violations.         |
```
LinZhihao-723: - `lint:cpp-static-check/fix` are not supported yet, right?
- We should also mention `lint:py-check/fix`

LinZhihao-723 (changes requested):
Thanks for the work! I've gone through all the files, and these are the comments I have so far.

Taskfile.yml
haiqi96: I remember that I have a discussion with Kirk on this and the conclusion is to not introduce too many extra changes. I am fine to apply the G_ prefix to lint-tasks.yml

README.md
haiqi96: lint:cpp-static-check/fix is actually supported. you can run it directly, it is just included in the lint:fix or lint:check flow.
I thought it would be helpful when you start to fix clang-tidy errors. Do you want me to change this behavior?

Taskfile.yml
LinZhihao-723: kk makes sense

lint-tasks.yml
haiqi96: .clang-format is the symlink for tools/yscope-dev-utils/lint-configs/.clang-format
I had a discussion with Kirk on this a while ago. Don't remember the details but I think the conclusion is that it's reasonable to use the symlink instead of directly using the source.

README.md
LinZhihao-723: I think let's remove it for now since it's not automated. Let me turn it on once all warnings are fixed

lint-tasks.yml
haiqi96: It's not a mistake. Depending on how you look at the dependencies. 

The line deps: [":config-cmake-project"] is added to ensure that although we don't run static check, the compilation database will still be generated.

Once we fix clang-tidy. config-cmake-project is then only needed by static check, so adding it as the dependency of cpp-static-check and cpp-static-fix should be good enough, and we don't need it to be a dependency of cpp-configs

You can also argue that you want the compilation database to be generated without having to run linter-fix, then maybe we should simply expose config-cmake-project to user (currently it's internal). The trade off is that now you need to manually run "task config-cmake-project" to generate the compilation database.

Imo, having it as a dependency of cpp-configs is a bit of hacky (and is a temporary solution), because cpp-configs is supposed to run only once to generate the symlink, but the compliation database should be generated whenever cmake file changes.

docs/conf/conf.py
haiqi96: very likely.

requirements-dev.txt
haiqi96: I and Kirk had a discussion about it a while ago. Can't find the chat anymore. 

Iirc, there are dependencies such as smart_open which is a build dependency, but requirements-lint.txt also needs it for python analysis. 

In an ideal world, we might want to have 3 requirements.txt. 1 for shared requirements. 1 for linter and 1 for dev dependencies but it would require changes to the yscope-dev-util.

Comparing to creating two requirements and duplicating the dependencies in both files, we decided it would be easier to just maintain one requirement file.

src/.clang-format
haiqi96: Sure

lint-tasks.yml
haiqi96: ditto of my previous comment on symlink and tools/yscope-dev-utils/lint-configs/.clang-format

Taskfile.yml
haiqi96: I think it is out of scope and would prefer to fix it in another PR, and I think it is low priority because we shouldn't expect people to change msgpack and GSL?

It will require some refactor so the task takes in a list of directory instead of just src/clp.

lint-tasks.yml
haiqi96: Sure.

requirements-dev.txt
LinZhihao-723: Sure

Taskfile.yml
LinZhihao-723: Yeah sure, I'm not suggesting we should do it in this PR, just to make sure we are aware of this behavior

lint-tasks.yml
LinZhihao-723: kk I see ur point. Make sense. Let's get back to this after clang-tidy fix

lint-tasks.yml
LinZhihao-723: Make sense

lint-tasks.yml
LinZhihao-723: I'm not saying we should use the config from `tools/yscope-dev-utils/lint-configs/.clang-format`; I'm saying we should explicitly say the `.clang-tidy` file is located under the root. Currently it's using the one from the working directory by default

haiqi96: Ah sorry that I misunderstood, sure.

lint-tasks.yml
haiqi96: I got 
`Usage: black [OPTIONS] SRC ...
Try 'black -h' for help.
Error: No such option: --output-format`

what's the version of black that you are using?

lint-tasks.yml
```diff
@@ -0,0 +1,213 @@
...
+          src/.clang-format \
+          Taskfile.yml
+
+  cmake:
+    internal: true
+    requires:
+      vars: ["FLAGS"]
+    cmd: |-
+      . "{{.LINT_VENV_DIR}}/bin/activate"
+      gersemi {{.FLAGS}} CMakeLists.txt {{.CLP_FFI_PY_SRC_DIR}}
```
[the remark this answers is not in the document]
haiqi96: Adding error reporting will introduce a lot of new changes to this PR. the intension is to mimic our existing linter flow so let's leave error handling in the future if we decide we need them.

lint-tasks.yml
```diff
@@ -0,0 +1,217 @@
...
+      vars: ["DOCFORMATTER_FLAGS", "BLACK_FLAGS", "RUFF_FLAGS", "SRC_DIRS"]
+    deps: ["venv"]
+    cmds:
+      - for:
+          var: "SRC_DIRS"
+        cmd: |-
+          . "{{.LINT_VENV_DIR}}/bin/activate"
+          docformatter {{.DOCFORMATTER_FLAGS}} --recursive {{.ITEM}}
+          black --color --line-length 100 {{.BLACK_FLAGS}} {{.ITEM}}
+          ruff check {{.RUFF_FLAGS}} {{.ITEM}}
```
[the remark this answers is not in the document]
haiqi96: @kirkrodrigues I think this could be helpful but it also adds a lot of lines. What's your opinion?

README.md
haiqi96: understood

lint-tasks.yml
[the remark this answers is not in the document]
kirkrodrigues: A command failure will already cause the task to stop immediately, right? And the command failure itself should print an error. So I'm not sure this suggestion adds anything? Unless I'm missing something.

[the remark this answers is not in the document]
haiqi96: depends on if you care about which command fails. There were some cases where I felt it was hard to figure out which command actually failed, but maybe not for this specific case.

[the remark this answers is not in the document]
haiqi96: well, let's not over complicate this and not use error handling for now. we can always add later if we figure

README.md
```diff
@@ -271,69 +277,65 @@ python -m unittest -bv
...
-    with their best effort to fix the reasonable ones.
+## Adding files
+Certain file types need to be added to our linting rules manually:
+
+- **CMake**. If adding a CMake file, add it (or its parent directory) as an argument to the
+  `gersemi` command in [lint-tasks.yaml](lint-tasks.yaml).
+  * If adding a directory, the file must be named `CMakeLists.txt` or use the `.cmake` extension.
+- **YAML**. If adding a YAML file (regardless of its extension), add it as an argument to the
+  `yamllint` command in [lint-tasks.yaml](lint-tasks.yaml).
+
```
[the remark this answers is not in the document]
haiqi96: Sure, sounds good.

Taskfile.yml
```diff
@@ -1,23 +1,47 @@
...
+  lint: "lint-tasks.yml"
+  utils: "tools/yscope-dev-utils/taskfiles/utils.yml"
 
 vars:
   # Paths
   BUILD_DIR: "{{.ROOT_DIR}}/build"
+  CLP_FFI_PY_CMAKE_BUILD_DIR: "{{.BUILD_DIR}}/cmake_build"
+  CLP_FFI_PY_CMAKE_CACHE: "{{.CLP_FFI_PY_CMAKE_BUILD_DIR}}/CMakeCache.txt"
+  CLP_FFI_PY_COMPILE_COMMANDS_DB: "{{.CLP_FFI_PY_CMAKE_BUILD_DIR}}/compile_commands.json"
+  CLP_FFI_PY_CPP_SRC_DIR: "{{.ROOT_DIR}}/src/clp_ffi_py"
```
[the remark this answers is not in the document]
haiqi96: @LinZhihao-723 How do you think? Maybe we can simply remove CLP_FFI_PY prefix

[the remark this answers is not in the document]
LinZhihao-723: hmmmm, I don't think removing the prefix would make it clearer. As a C-extension project, having a cmake build directory for the Python package should be expected, or it's actually considered as a part of the Python object. Let's keep the prefix unless you have a strong disagreement

docs/conf/conf.py
LinZhihao-723: Can we revert this change since it will no longer be formatted with the latest lint task config?

haiqi96:
```suggestion
```

LinZhihao-723 (approved):
For PR title, how about:
```
feat(linting): Add Taskfile tasks to run linters.
```

PR #77, opened 2024-10-27
Rename `Deserializer` to `FourByteDeserializer` to be consistent with the serialization APIs.

LinZhihao-723:
# References

# Description

This PR depends on #73 
This PR renames `Deserializer` to `FourByteDeserializer` to be consistent with the current serialization APIs (which is called `FourByteSerializer`). Similarly, we rename `PyDeserializer` to `PyFourByteDeserializer` in ffi.
After this PR, we will have a cleaner namespace for the coming key-value pair IR format, which will use `Serializer` and `Deserializer`  without any prefix as the names for write/read APIs.
Notice that `DeserializationBuffer` is not renamed, as we will reuse this class in the key-value pair IR format.

# Validation performed

- Ensure all unit tests passed
- Ensure the doc site builds correctly with all renamed APIs reflected

## Release Notes

- **New Features**
  - Introduced the `FourByteDeserializer` class, enhancing data deserialization capabilities.
  - Updated public API to reflect the new deserializer, replacing `Deserializer` with `FourByteDeserializer`.

- **Bug Fixes**
  - Improved error handling in log reading methods to ensure robustness with the new deserializer.

- **Tests**
  - Refactored test cases to align with the new `FourByteDeserializer`, ensuring comprehensive coverage of its functionality.

These updates enhance the overall performance and reliability of the deserialization process within the application.


src/clp_ffi_py/modules/ir_native.cpp
```diff
@@ -1,7 +1,7 @@
 #include <clp_ffi_py/Python.hpp>  // Must always be included before any other header files
 
-#include <clp_ffi_py/ir/native/PyDeserializer.hpp>
 #include <clp_ffi_py/ir/native/PyDeserializerBuffer.hpp>
+#include <clp_ffi_py/ir/native/PyFourByteDeserializer.hpp>
```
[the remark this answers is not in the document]
LinZhihao-723: Things inside PyDeserializerBuffer.* should be fine

CMakeLists.txt
```diff
@@ -0,0 +1,137 @@
...
+# this project.
+set(CLP_FFI_PY_CLP_CORE_SOURCES
+    ${CLP_CORE_ROOT}/src/clp/BufferReader.cpp
+    ${CLP_CORE_ROOT}/src/clp/ffi/ir_stream/decoding_methods.cpp
+    ${CLP_CORE_ROOT}/src/clp/ffi/ir_stream/encoding_methods.cpp
+    ${CLP_CORE_ROOT}/src/clp/ffi/ir_stream/utils.cpp
+    ${CLP_CORE_ROOT}/src/clp/ffi/encoding_methods.cpp
+    ${CLP_CORE_ROOT}/src/clp/ir/parsing.cpp
+    ${CLP_CORE_ROOT}/src/clp/ReaderInterface.cpp
+)
```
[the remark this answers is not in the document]
LinZhihao-723: Not related in this PR. Will be merged in #73

src/clp_ffi_py/ir/native/PyFourByteDeserializer.cpp
```diff
@@ -61,43 +61,44 @@ PyMethodDef PyDeserializer_method_table[]{
 
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays)
 PyDoc_STRVAR(
-        cPyDeserializerDoc,
-        "Namespace for all CLP IR deserialization methods.\n\n"
+        cPyFourByteDeserializerDoc,
+        "Namespace for all CLP four byte IR deserialization methods.\n\n"
```
haiqi96:
```suggestion
        "Namespace for all CLP four-byte encoded IR deserialization methods.\n\n"
```

So it's consistent with what you wrote in the files below

tests/test_ir/test_deserializer.py
```diff
@@ -225,7 +225,7 @@ def test_deserializer_with_random_logs(self) -> None:
             )
 
 
-class TestCaseDeserializerDecompress(TestCaseDeserializerBase):
+class TestCaseFourByteDeserializerDecompress(TestCaseFourByteDeserializerBase):
```
haiqi96, line 228: Nit (and you can ignore this if you don't agree with me)
The names look terribly long for me in this file. It should be clear that if the base class is four bytes, then inherited class will also be testing four bytes.
Do we have a plan to support 8bytes in the future? If not, maybe we can remove "FourByte" from the inhertied class. If we have such a plan, maybe it would worth splitting the test file into 2, one for 4 bytes and one for 8 bytes?

haiqi96 (dismissed):
Looks good to me, the review is straight forward since the PR only consists of renamings.

Added two nit comments. feel free to apply them or ignore

tests/test_ir/test_deserializer.py
LinZhihao-723: no, we don't have plan to support 8byte...
We will rewrite most of these functions for v2 format. Once that format is formally released, all v1 related code will be removed from the code base.
I will pay more attentions to test case namings for v2 unit tests :)

haiqi96 (approved):
lgtm

PR #81, opened 2024-11-03
feat: Add Python class `KeyValuePairLogEvent` as a wrapper of `clp::ffi::KeyValuePairLogEvent`.

LinZhihao-723:
# References

# Description

This PR introduces a new Python class, `KeyValuePairLogEvent`, that is implemented in Python C-extension. it is a wrapper of `clp::ffi::KeyValuePairLogEvent`. This class contains a method `to_dict` to serialize the underlying key-value pairs into a Python dictionary. It is not designed to be directly instantiated by users, but to receive results emitted from CLP key-value pair IR format's deserializer. However, this class can be instantiated directly from a Python dictionary for testing purposes despite it being very inefficient: for every input dictionary, we serialize it into a new CLP key-value pair IR stream and deserialize it into a key-value pair log event (this is the only option before we have a way to construct a KeyValuePairLogEvent from an arbitrary map-like data structure directly). In the unit tests, we test the conversion between a Python dictionary and a `KeyValuePairLogEvent` instance, ensuring accurate serialization and deserialization in both directions. The test inputs are open-source JSON logs from elasticsearch and spark (downloaded from CLP's testing dataset [here](https://docs.yscope.com/clp/main/user-guide/resources-datasets))

Notice that the conversion from `KeyValuePairLogEvent` to Python dictionary in this PR is implemented in a very inefficient way: it first serializes the underlying key-value pairs into a JSON string and then parses the JSON string using Python utils into a dictionary. The efficient way to implement `to_dict` is to construct the dictionary directly inside the C-extension layer using the similar algorithm in `clp::ffi::KeyValuePairLogEvent::serialize_to_json`. However, we will defer this to a future PR since the size of this PR is already large enough; it's more helpful if we can build some end-to-end testing before optimizing for performance.

This PR introduces two additional C++ dependencies: `msgpack` and `GSL`. `msgpack` is required by CLP's key-value pair IR format serializer/deserializer. `GSL` is to declare ownership of the dynamically allocated memory inside a Python object. Notice that both dependencies are installed as git submodules, and included in cmake using `add_subdirectory`. We also apply `EXCLUDE_FROM_ALL` to `msgpack` to not install it on the built Python wheels. The current way of dependency management is still considered a temporary solution. We may revisit dependency management in the near future.

To address the clang-tidy issues about not-directly-included headers in the new files, this PR also adds a section in `Python.hpp` to export all Python headers used in the project (using `IWYU pragma`). These header exports are only added for lining purpose, so we add a cmake-specified macro to exclude them when building the actual Python C-extension library.

# Validation performed

- Ensure unit tests all passed.
- Ensure workflow passed.

## Release Notes

- **New Features**
	- Introduced `KeyValuePairLogEvent` class for managing log events as key-value pairs.
	- Added `parse_json_str` function for converting JSON strings to Python objects.
	- New `unpack_msgpack` function for handling MessagePack data.
	- Enhanced build configuration with new submodules for `msgpack` and `GSL`.
	- Added a macro for C linkage in Python APIs.

- **Bug Fixes**
	- Enhanced error handling in serialization and deserialization processes.

- **Tests**
	- Added tests for `KeyValuePairLogEvent` functionality.
	- Introduced `JsonLinesFileReader` for reading JSON line files.

- **Documentation**
	- Updated configuration files to improve build and linting processes.


tests/test_ir/test_key_value_pair_log_event.py
```diff
@@ -0,0 +1,31 @@
...
+        test_src_dir: Path = current_dir / Path(TestCaseKeyValuePairLogEvent.input_src_dir)
+        for file_path in test_src_dir.rglob("*"):
+            if not file_path.is_file():
+                continue
+            json_file_reader: JsonFileReader = JsonFileReader(file_path)
+            for expected in json_file_reader.read_json_lines():
+                self.assertEqual(isinstance(expected, dict), True)
+                actual: KeyValuePairLogEvent = KeyValuePairLogEvent(expected)
+                serialized_py_dict: Dict[Any, Any] = actual.to_dict()
+                self.assertEqual(expected, serialized_py_dict)
```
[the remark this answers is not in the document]
LinZhihao-723: Add `self.assertIsInstance(expected, dict, "Input must be a dictionary")`

tests/test_ir/test_utils.py
```diff
@@ -22,6 +24,32 @@
...
+        self.file_path: Path = file_path
+
+    def read_json_lines(self) -> Generator[Any, None, None]:
+        """
+        Reads each line in the JSON file, parses it as a JSON object, and yields the JSON object.
+
+        :yield: Parsed JSON object for each line in the file.
+        """
+        with open(self.file_path, "r", encoding="utf-8") as file:
+            yield from (json.loads(line.strip()) for line in file)
```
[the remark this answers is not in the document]
LinZhihao-723: This is only used in unit tests. If error occurs, just let it throw.

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -0,0 +1,342 @@
...
+    return 0;
+}
+
+CLP_FFI_PY_METHOD auto PyKeyValuePairLogEvent_dealloc(PyKeyValuePairLogEvent* self) -> void {
+    self->clean();
+    PyObject_Del(self);
+}
+
+CLP_FFI_PY_METHOD auto PyKeyValuePairLogEvent_to_dict(PyKeyValuePairLogEvent* self) -> PyObject* {
+    // TODO: use an efficient algorithm to turn the underlying log event to a Python dictionary
```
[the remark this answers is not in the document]
LinZhihao-723: It's added there intentionally and waiting for future PRs.

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -0,0 +1,342 @@
...
+    }
+    auto const json_str{serialized_json_result.value().dump()};
+    auto* parsed_json{py_utils_parse_json_str(json_str)};
+    if (nullptr == parsed_json) {
+        return nullptr;
+    }
+    if (false == static_cast<bool>(PyDict_Check(parsed_json))) {
+        PyErr_SetString(PyExc_TypeError, "Serialized JSON object is not a dictionary");
+    }
+    return parsed_json;
```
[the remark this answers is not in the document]
LinZhihao-723: right right

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.hpp
```diff
@@ -0,0 +1,73 @@
...
+    auto default_init() -> void { m_kv_pair_log_event = nullptr; }
+
+    /**
+     * Releases the memory allocated for underlying data fields.
+     */
+    auto clean() -> void { delete m_kv_pair_log_event; }
+
+    [[nodiscard]] auto get_kv_pair_log_event() const -> clp::ffi::KeyValuePairLogEvent const& {
+        return *m_kv_pair_log_event;
+    }
```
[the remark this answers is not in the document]
LinZhihao-723: Will switch to return a pointer instead to be consistent with other classes

clp_ffi_py/utils.py
```diff
@@ -45,3 +47,13 @@ def serialize_dict_to_msgpack(dictionary: Dict[Any, Any]) -> bytes:
...
+
+
+def parse_json_str(json_str: str) -> Any:
+    """
+    Wrapper of `json.loads`, which parses a JSON string into a Python object.
+
+    :param json_str: The JSON string to parse.
+    :return: The parsed JSON object.
+    """
+    return json.loads(json_str)
```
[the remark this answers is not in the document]
LinZhihao-723: If an error happens, we will handle it inside C-extension code.

tests/test_ir/test_key_value_pair_log_event.py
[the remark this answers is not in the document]
LinZhihao-723: We can't do negative tests yet. Waiting for some updates in clp oss.

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -0,0 +1,337 @@
...
+        return std::nullopt;
+    }
+
+    return std::move(ir_unit_handler.m_log_event);
+}
+
+CLP_FFI_PY_METHOD auto PyKeyValuePairLogEvent_dealloc(PyKeyValuePairLogEvent* self) -> void {
+    self->clean();
+    PyObject_Del(self);
+}
```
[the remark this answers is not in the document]
LinZhihao-723: From this doc I think u're correct:
https://docs.python.org/3/extending/newtypes_tutorial.html
Will push with the coming commits

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.hpp
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * Initializes the pointers to nullptr by default. Should be called once the object is
+     * allocated.
+     */
+    auto default_init() -> void { m_kv_pair_log_event = nullptr; }
+
+    /**
+     * Releases the memory allocated for underlying data fields.
+     */
+    auto clean() -> void {
```
haiqi96, line 51: nit. how about dealloc(), or deinit()
When I see clean, it reminds me of clear()

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -0,0 +1,340 @@
...
+        return -1;
+    }
+
+    if (false == static_cast<bool>(PyDict_Check(dictionary))) {
+        PyErr_SetString(PyExc_TypeError, "`dictionary` must be a Python dictionary object");
+        return -1;
+    }
+    PyDictObject* py_dict{py_reinterpret_cast<PyDictObject>(dictionary)};
+
+    auto optional_kv_pair_log_event{convert_py_dict_to_key_value_pair_log_event(py_dict)};
```
haiqi96, line 190: nit: how about `kv_pair_log_event`?

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -0,0 +1,340 @@
...
+        }
+        auto const ir_unit_type{result.value()};
+        if (ir_unit_type == clp::ffi::ir_stream::IrUnitType::SchemaTreeNodeInsertion) {
+            continue;
+        }
+        if (ir_unit_type != clp::ffi::ir_stream::IrUnitType::LogEvent) {
+            PyErr_SetString(PyExc_RuntimeError, "Unexpected Ir Unit Type");
+            return std::nullopt;
+        }
+        break;
```
haiqi96:
```suggestion
        if (clp::ffi::ir_stream::IrUnitType::LogEvent == ir_unit_type) {
            break;
        }
        if (ir_unit_type != clp::ffi::ir_stream::IrUnitType::SchemaTreeNodeInsertion) {
            PyErr_SetString(PyExc_RuntimeError, "Unexpected Ir Unit Type");
            return std::nullopt;
        }
```

haiqi96: I feel this can be simplified? Hope I don't miss any corner case

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -0,0 +1,340 @@
...
+        return false;
+    }
+    return true;
+}
+
+auto PyKeyValuePairLogEvent::get_py_type() -> PyTypeObject* {
+    return m_py_type.get();
+}
+
+auto PyKeyValuePairLogEvent::module_level_init(PyObject* py_module) -> bool {
```
haiqi96, line 332: I assume module_level_init will always get called before get_py_type by user end?

haiqi96 (commented):
First batch of review

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
LinZhihao-723: Yes

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
LinZhihao-723: It's our new standard: we should add `optional_` for returned optional variable

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.hpp
LinZhihao-723: This is following the standard we made for Python object classes. `init`, `default_init`, `clean`, and some other static members are the same for all Python objects. I think `deinit` might be a better name though, but shall we only apply these to newly added classes? @kirkrodrigues what do u think?

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
LinZhihao-723: Right correct... sorry

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -0,0 +1,340 @@
...
+                    result.error().message().c_str()
+            );
+            return std::nullopt;
+        }
+        auto const ir_unit_type{result.value()};
+        if (clp::ffi::ir_stream::IrUnitType::LogEvent == ir_unit_type) {
+            break;
+        }
+        if (clp::ffi::ir_stream::IrUnitType::SchemaTreeNodeInsertion != ir_unit_type) {
+            continue;
```
haiqi96: not sure why this `continue` is here, maybe git messed something up or maybe I made a mistake when adding the suggestion.

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
haiqi96: kk, sounds good

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.hpp
kirkrodrigues: Since the existing code is consistent, I would prefer to maintain consistency and then address the rename in a single PR.

Btw, didn't we consider deinit before?

haiqi96: sure, doesn't need to be in this PR.

tests/test_ir/test_utils.py
```diff
@@ -22,6 +24,32 @@
 from clp_ffi_py.wildcard_query import WildcardQuery
 
 
+class JsonFileReader:
```
haiqi96: Nit, it seems there is a specific name for Json files with each line as a Json object `JSON Lines`(.jsonl). https://jsonlines.org/
If you want, we can change the Readername to be JsonLineFileReader, and read_json_lines -> read_json_values?

Don't think it has to be in this change, but just wonder if we want to consider using `Json Lines`. It might cause some confusion for our customers though
@kirkrodrigues

kirkrodrigues: How about:

* `JsonFileReader` -> `JsonLinesFileReader`
* `read_json_lines` -> `read_line`
  * I don't think this should be plural since it yields after each line, right?

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
LinZhihao-723: Yeah it's my bad... sorry, removed

tests/test_ir/test_utils.py
haiqi96: it's yield from. so it actually yields a list of lines. I feel it should be plural

haiqi96 (dismissed):
The PR looks good to me. I will leave it up to Zhihao to decide if he wants to update the JsonFileReader to JsonLineFileReader

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -0,0 +1,339 @@
...
+    }
+    auto const json_str{serialized_json_result.value().dump()};
+    auto* parsed_json{py_utils_parse_json_str(json_str)};
+    if (nullptr == parsed_json) {
+        return nullptr;
+    }
+    if (false == static_cast<bool>(PyDict_Check(parsed_json))) {
+        PyErr_SetString(PyExc_TypeError, "Serialized JSON object is not a dictionary");
+        return nullptr;
+    }
```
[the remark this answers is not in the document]
LinZhihao-723: Make sense, my bad

tests/test_ir/test_utils.py
LinZhihao-723: 1. Applied. I think the suggestions make sense.
2. It should not be plural as the generator returns object one by one. My bad.

tests/test_ir/test_key_value_pair_log_event.py
```diff
@@ -0,0 +1,31 @@
...
+        """
+        Tests the conversion between a Python dictionary and a `KeyValuePairLogEvent` instance,
+        ensuring accurate serialization and deserialization in both directions.
+        """
+        current_dir: Path = Path(__file__).resolve().parent
+        test_src_dir: Path = current_dir / Path(TestCaseKeyValuePairLogEvent.input_src_dir)
+        for file_path in test_src_dir.rglob("*"):
+            if not file_path.is_file():
+                continue
+            json_file_reader: JsonLinesFileReader = JsonLinesFileReader(file_path)
```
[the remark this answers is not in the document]
haiqi96: I kind of agree with this review. maybe we should at least catch the case when there's no file under test_src_dir which will cause the test to automatically pass.

CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ if(DEFINED SKBUILD_PROJECT_NAME)
         "The version of the project parsed by scikit-build-core."
         FORCE
     )
+    set(CLP_FFI_PY_LINT OFF CACHE BOOL "Disable linting." FORCE)
```
kirkrodrigues: Could rename this to `CLP_FFI_PY_ENABLE_LINTING` and then you wouldn't need the description, right?

CMakeLists.txt
```diff
@@ -59,20 +61,32 @@ set(CLP_FFI_PY_CLP_CORE_DIR "${CLP_FFI_PY_SRC_DIR}/clp/components/core")
 
 add_subdirectory(${CLP_FFI_PY_CLP_CORE_DIR}/src/clp/string_utils)
 
+set(MSGPACK_CXX20 ON CACHE BOOL "Enable C++20 in msgpack")
+set(MSGPACK_USE_BOOST OFF CACHE BOOL "Disable Boost in msgpack")
```
kirkrodrigues: I think we should add `FORCE` to these to override the cache? Or are you anticipating a use case where we would want to set something different externally?

CMakeLists.txt
```diff
@@ -59,20 +61,32 @@ set(CLP_FFI_PY_CLP_CORE_DIR "${CLP_FFI_PY_SRC_DIR}/clp/components/core")
...
+add_subdirectory(${CLP_FFI_PY_SRC_DIR}/msgpack EXCLUDE_FROM_ALL)
+
+add_subdirectory(${CLP_FFI_PY_SRC_DIR}/GSL)
```
kirkrodrigues, line 66: Do you want to move this before msgpack so that it's alphabetical?

.clang-format
```diff
@@ -77,9 +77,7 @@ IncludeCategories:
   - Regex: "^<clp_ffi_py"
     Priority: 4
   # Third-party headers. Update when adding new third-party library
-  - Regex: "^<(clp)"
-    Priority: 3
-  - Regex: "<(json)"
+  - Regex: "<(clp|json|msgpack|outcome)"
```
kirkrodrigues:
```suggestion
  - Regex: "<(clp|gsl|json|msgpack|outcome)"
```

tests/test_ir/test_json/elasticsearch.jsonl
kirkrodrigues, line 1: Can we delete the unnecessary spaces in this file?

src/msgpack
kirkrodrigues, line 1: It looks like this is using a commit one past the last release? Any reason why?

clp_ffi_py/utils.py
```diff
@@ -1,7 +1,9 @@
+import json
 from datetime import datetime, tzinfo
 from typing import Any, Dict, Optional
 
 import dateutil.tz
+
 import msgpack
```
kirkrodrigues: `ruff` is treating `msgpack` as a [first-party](https://docs.astral.sh/ruff/settings/#lint_isort_section-order) import rather than a third-party one. To prevent this, we need to exclude `src` from the directories `ruff` [considers](https://docs.astral.sh/ruff/settings/#src).

Add to `[tool.ruff]` in `pyproject.toml`:
```toml
src = ["."]
```

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -0,0 +1,339 @@
...
+    ) -> IRErrorCode {
+        return IRErrorCode::IRErrorCode_Success;
+    }
+
+    [[nodiscard]] static auto handle_end_of_stream() -> IRErrorCode {
+        return IRErrorCode::IRErrorCode_Success;
+    }
+
+    // NOLINTNEXTLINE(*)
+    std::optional<clp::ffi::KeyValuePairLogEvent> m_log_event;
```
kirkrodrigues: * Why are turning off linting entirely?
* Why is this a public member?

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.hpp
```diff
@@ -0,0 +1,88 @@
...
+#include <gsl/gsl>
+
+#include <clp/ffi/KeyValuePairLogEvent.hpp>
+
+#include <clp_ffi_py/PyObjectUtils.hpp>
+
+namespace clp_ffi_py::ir::native {
+/**
+ * A PyObject structure functioning as a Python-compatible interface to retrieve a key-value pair
+ * log event. The underlying data is pointed by `m_kv_pair_log_event`.
```
kirkrodrigues:
```suggestion
 * log event. The underlying data is pointed to by `m_kv_pair_log_event`.
```

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.hpp
```diff
@@ -0,0 +1,88 @@
...
+ */
+class PyKeyValuePairLogEvent {
+public:
+    // Delete default constructor to disable direct instantiation.
+    PyKeyValuePairLogEvent() = delete;
+
+    // Delete copy/move constructors and assignments
```
kirkrodrigues:
```suggestion
    // Delete copy/move constructors and assignment operators
```

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.hpp
kirkrodrigues: https://github.com/y-scope/clp-ffi-py/pull/3#discussion_r1267397537

src/clp_ffi_py/Python.hpp
```diff
@@ -8,3 +8,20 @@
 #define PY_SSIZE_T_CLEAN
 #include <Python.h>
 #include <structmember.h>
```
kirkrodrigues, line 10: Do you remember why we included `structmember.h` in this file? Was it to try and resolve a clang-tidy warning?

src/clp_ffi_py/Python.hpp
```diff
@@ -8,3 +8,20 @@
...
+
+// clang-format off
+#ifdef CLP_FFI_PY_ENABLE_LINTING
+// The following headers are added to export Python headers to get rid of clang-tidy warnings.
```
kirkrodrigues:
```suggestion
// Inform IWYU of the headers that we use that are exported by Python.h
```

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -0,0 +1,339 @@
...
+using clp::ffi::ir_stream::IRErrorCode;
+
+namespace clp_ffi_py::ir::native {
+namespace {
+/**
+ * Class that implements `clp::ffi::ir_stream::IrUnitHandlerInterface` for deserializing log events.
+ */
+class IrUnitHandler {
+public:
+    // Implements `clp::ffi::ir_stream::IrUnitHandlerInterface` interface
```
kirkrodrigues:
```suggestion
    // Methods that implement the `clp::ffi::ir_stream::IrUnitHandlerInterface` interface
```

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -0,0 +1,339 @@
...
+ * Callback of `PyKeyValuePairLogEvent`'s `__init__` method.
+ */
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays)
+PyDoc_STRVAR(
+        cPyKeyValuePairLogEventDoc,
+        "This class represents a key-value pair log event and provides ways to access the"
+        " underlying log data. This class is designed to be instantiated by the IR deserializer."
+        " However, direct instantiation using the `__init__` method is also supported for testing"
+        " purposes, although this may not be as efficient as emitting from the IR deserializer.\n\n"
+        "The signature of `__init__` method is shown as following:\n\n"
```
kirkrodrigues:
```suggestion
        " purposes, although this may not be as efficient as emission from the IR deserializer.\n\n"
        "The signature of `__init__` method is:\n\n"
```

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -0,0 +1,339 @@
...
+        "__init__(self, dictionary)\n\n"
+        "Initializes an object that represents a key-value pair log event from the given Python"
+        " dictionary. Notice that each object should be strictly initialized only once. Double"
+        " initialization will result in memory leak.\n\n"
```
kirkrodrigues:
```suggestion
        " dictionary. Note that each object should only be initialized once. Double initialization"
        " will result in a memory leak.\n\n"
```

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -0,0 +1,339 @@
...
+        ":param dictionary: A dictionary representing the key-value log event, where all keys are"
+        " expected to be of string type, including keys inside any sub-dictionaries.\n"
```
kirkrodrigues:
```suggestion
        " expected to strings, including keys inside any sub-dictionaries.\n"
```

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -0,0 +1,339 @@
...
+    auto optional_kv_pair_log_event{convert_py_dict_to_key_value_pair_log_event(py_dict)};
+    if (false == optional_kv_pair_log_event.has_value()) {
+        return -1;
+    }
+
+    return self->init(std::move(optional_kv_pair_log_event.value())) ? 0 : -1;
+}
+
+CLP_FFI_PY_METHOD auto PyKeyValuePairLogEvent_to_dict(PyKeyValuePairLogEvent* self) -> PyObject* {
+    // TODO: use an efficient algorithm to turn the underlying log event to a Python dictionary
```
kirkrodrigues:
```suggestion
    // TODO: Use an efficient algorithm to turn the underlying log event into a Python dictionary
```

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -0,0 +1,339 @@
...
+};
+
+/**
+ * Converts the given Python dictionary to `clp::ffi::KeyValuePairLogEvent`.
+ * NOTE: Currently, there is no direct utility for constructing a `clp::ffi::KeyValuePairLogEvent`
+ * from a dictionary-like structure. As a workaround, this function serializes the dictionary
+ * using the key-value pair IR format, then deserializes it to create a `KeyValuePairLogEvent`
+ * instance. This approach is inefficient and intended solely for testing purposes, as it allows
+ * instance creation without a full IR stream. Future implementations should replace this method
+ * with a more efficient conversion once a direct utility is available.
```
kirkrodrigues:
```suggestion
 * instance creation without a full IR stream. TODO: Replace this method with a more efficient
 * conversion once a direct utility is available.
```

I think it makes more sense to put the TODO here (and get rid of the one inside the implementation) so we don't forget to also update the docstring.

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.hpp
```diff
@@ -0,0 +1,88 @@
...
+    auto operator=(PyKeyValuePairLogEvent&&) -> PyKeyValuePairLogEvent& = delete;
+
+    // Destructor
+    ~PyKeyValuePairLogEvent() = default;
+
+    /**
+     * Initializes the underlying data with the given inputs. Since the memory allocation of
+     * `PyKeyValuePairLogEvent` is handled by CPython's allocator, cpp constructors will not be
+     * explicitly called. This function serves as the default constructor initialize the underlying
+     * key-value pair log event. It has to be manually called whenever creating a new
```
kirkrodrigues:
```suggestion
     * explicitly called. This function serves as the default constructor to initialize the
     * underlying key-value pair log event. It has to be called manually to create a
```

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.hpp
```diff
@@ -0,0 +1,88 @@
...
+     * dynamically created and initialized during the execution of `module_level_init`.
+     * @return Python type object associated with `PyKeyValuePairLogEvent`.
+     */
+    [[nodiscard]] static auto get_py_type() -> PyTypeObject*;
+
+    /**
+     * Creates and initializes `PyKeyValuePairLogEvent` as a Python type, and then incorporates this
+     * type as a Python object into the py_module module.
+     * @param py_module This is the Python module where the initialized `PyKeyValuePairLogEvent`
+     * will be incorporated.
```
kirkrodrigues:
```suggestion
     * @param py_module The Python module where the initialized `PyKeyValuePairLogEvent` will be
     * incorporated.
```

src/clp_ffi_py/utils.hpp
```diff
@@ -60,6 +63,15 @@ auto get_py_bool(bool is_true) -> PyObject*;
 template <typename int_type>
 auto parse_py_int(PyObject* py_int, int_type& val) -> bool;
 
+/**
+ * Unpacks the given msgpack byte sequence.
+ * @param msgpack_byte_sequence
+ * @return A result containing the unpack msgpack object handle on success or an error string
```
kirkrodrigues:
```suggestion
 * @return A result containing the unpacked msgpack object handle on success or an error string
```

src/clp_ffi_py/Python.hpp
```diff
@@ -8,3 +8,20 @@
...
+// IWYU pragma: begin_exports
+#include <bytesobject.h>
+#include <dictobject.h>
+#include <methodobject.h>
+#include <modsupport.h>
+#include <object.h>
+#include <objimpl.h>
+#include <pyerrors.h>
+#include <pymacro.h>
+#include <typeslots.h>
```
kirkrodrigues, line 24: I haven't verified that these are all necessary. Let's do that in #76.

src/clp_ffi_py/api_decoration.hpp
```diff
@@ -0,0 +1,11 @@
+#ifndef CLP_FFI_PY_API_DECORATION_HPP
+#define CLP_FFI_PY_API_DECORATION_HPP
+
+/**
+ * `CLP_FFI_PY_METHOD` should be added at the beginning of a function's declaration/implementation
+ * to decorate any APIs that are directly invoked by Python's interpreter. The macro expands to
+ * `extern "C"` to ensure C linkage.
+ */
+#define CLP_FFI_PY_METHOD extern "C"
```
kirkrodrigues, line 9: Do we need to add this to any existing methods?

tests/test_ir/test_utils.py
```diff
@@ -22,6 +24,33 @@
 from clp_ffi_py.wildcard_query import WildcardQuery
 
 
+class JsonLinesFileReader:
+    """
+    Class for reading JSON files.
+
+    It assumes each line in the file is a JSON string, and the parser parses each line into a JSON
+    object, and return then through a generator.
```
kirkrodrigues:
```suggestion
    object, and returns then through a generator.
```

tests/test_ir/test_utils.py
```diff
@@ -22,6 +24,33 @@
...
+
+        :param file_path: Path to the JSON file to read.
+        """
+        self.file_path: Path = file_path
+
+    def read_lines(self) -> Generator[Any, None, None]:
+        """
+        Reads each line in the JSON file, parses it as a JSON object, and yields the JSON object.
+
+        :yield: Parsed JSON object for each line in the file.
```
kirkrodrigues, line 47: Is `yield` a recognized docstring tag?

kirkrodrigues (changes requested):

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
LinZhihao-723: This should be a simple struct and there's no need to hide the log event IMO to make the code simple. Turning off the linting because we don't have the latest clang-tidy added to allow simple struct.

src/clp_ffi_py/Python.hpp
LinZhihao-723: I don't remember the exact reason, but it wasn't clang-tidy related

src/clp_ffi_py/api_decoration.hpp
LinZhihao-723: No. For the current methods, we used `extern "C"` to wrap the entire code block that contains C-linkage methods. As we discussed before we should apply `extern "C"` method by method. This macro is mirroring what we added in Go-ffi (even though in Go ffi the usage is a little more complicated)

src/clp_ffi_py/Python.hpp
LinZhihao-723: I'm ok to defer them to other PRs, but just wanna make an agreement that this (IWYU + macro) will be our solution to resolve clang-tidy warnings

CMakeLists.txt
LinZhihao-723: sure

CMakeLists.txt
LinZhihao-723: yeah using `FORCE` makes sense

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.hpp
LinZhihao-723: It this a typo or?

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
kirkrodrigues: Sure, so then we should:

* name it `log_event` (our style guide takes precedence over incorrectly configured linting).
* disable only the linting check(s) that is(are) being violated.
* Add a TODO to update the turned off linting when we add the correct linting setting.

Right now, this statement can easily be forgotten.

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.hpp
kirkrodrigues: No, that's the grammatically correct thing to write. When X points at Y, you write "Y is pointed to by X" (or "Y is pointed at by X").

src/clp_ffi_py/Python.hpp
kirkrodrigues: Everything seems to compile if we remove it. Let's revisit in #76.

src/clp_ffi_py/Python.hpp
kirkrodrigues: As we discussed offline, I'm not suggesting we remove this from this PR. I'm saying we can review it in #76.

src/clp_ffi_py/Python.hpp
LinZhihao-723: `structmember.h` seems to define some macros for CPython. I think we included this header for some safety concern (i.e., ensure some macro is defined), but I don't remember exactly which part of the code requires it. But yeah, let's revisit it in future PRs.

tests/test_ir/test_utils.py
LinZhihao-723: Yes. We have `:yield:` in `clp_ffi_py.ir.readers`, and here's the generated doc with `yield`: https://docs.yscope.com/clp-ffi-py/main/api/clp_ffi_py.ir.html#clp_ffi_py.ir.ClpIrStreamReader.search

tests/test_ir/test_json/elasticsearch.jsonl
LinZhihao-723: Done

src/msgpack
LinZhihao-723: Should be a mistake that I didn't check out to a particular release and leave it at the latest commit.
Changed it to v7.0.0

clp_ffi_py/utils.py
LinZhihao-723: thx for catching this. Didn't notice this difference is caused by ruff treating msgpack as first-party import

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.hpp
```diff
@@ -0,0 +1,88 @@
...
+     * Releases the memory allocated for underlying data fields.
+     */
+    auto clean() -> void {
+        delete m_kv_pair_log_event;
+        m_kv_pair_log_event = nullptr;
+    }
+
+    [[nodiscard]] auto get_kv_pair_log_event() const -> clp::ffi::KeyValuePairLogEvent const* {
+        return m_kv_pair_log_event;
+    }
```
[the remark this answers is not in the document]
LinZhihao-723: we don't check this because we assume invalid objects will never be created, assuming CPython system runs correctly

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.hpp
```diff
@@ -0,0 +1,87 @@
...
+     * Releases the memory allocated for underlying data fields.
+     */
+    auto clean() -> void {
+        delete m_kv_pair_log_event;
+        m_kv_pair_log_event = nullptr;
+    }
+
+    [[nodiscard]] auto get_kv_pair_log_event() const -> clp::ffi::KeyValuePairLogEvent const* {
+        return static_cast<clp::ffi::KeyValuePairLogEvent const*>(m_kv_pair_log_event);
+    }
```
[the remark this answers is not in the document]
LinZhihao-723: Same

tests/test_ir/test_key_value_pair_log_event.py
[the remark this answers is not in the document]
LinZhihao-723: Changed

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -0,0 +1,341 @@
...
+        " underlying log data. This class is designed to be instantiated by the IR deserializer."
+        " However, direct instantiation using the `__init__` method is also supported for testing"
+        " purposes, although this may not be as efficient as emission from the IR deserializer.\n\n"
+        "The signature of `__init__` method is:\n\n"
+        "__init__(self, dictionary)\n\n"
+        "Initializes an object that represents a key-value pair log event from the given Python"
+        " dictionary. Note that each object should only be initialized once. Double initialization"
+        " will result in a memory leak.\n\n"
+        ":param dictionary: A dictionary representing the key-value log event, where all keys are"
+        " expected to strings, including keys inside any sub-dictionaries.\n"
```
kirkrodrigues:
```suggestion
        "This class represents a key-value pair log event and provides methods to access the"
        " key-value pairs. This class is designed to be instantiated by the IR deserializer."
        " However, direct instantiation using the `__init__` method is also supported for testing"
        " purposes, although this may not be as efficient as emission from the IR deserializer.\n\n"
        "__init__(self, dictionary)\n\n"
        "Initializes a :class:`KeyValuePairLogEvent` from the given Python dictionary. Note that"
        " each object should only be initialized once. Double initialization will result in a"
        " memory leak.\n\n"
        ":param dictionary: A dictionary representing the key-value pair log event, where all keys"
        " must be strings, including keys inside any sub-dictionaries.\n"
```

kirkrodrigues: Some new edits based on suggestions in #86.

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -0,0 +1,341 @@
...
+/**
+ * Callback of `PyKeyValuePairLogEvent`'s `to_dict` method.
+ */
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays)
+PyDoc_STRVAR(
+        cPyKeyValuePairLogEventToDictDoc,
+        "to_dict(self)\n"
+        "--\n\n"
+        "Converts the underlying key-value pair log event into a Python dictionary.\n\n"
+        ":return: Serialized log event in a Python dictionary.\n"
```
kirkrodrigues:
```suggestion
        "Converts the log event into a Python dictionary.\n\n"
        ":return: The log event as a Python dictionary.\n"
```

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
kirkrodrigues: Ditto [comment](https://github.com/y-scope/clp-ffi-py/pull/86#discussion_r1841972451) about documenting types.

CMakeLists.txt
```diff
@@ -57,22 +59,37 @@ set(CLP_FFI_PY_SRC_DIR "${CMAKE_CURRENT_SOURCE_DIR}/src")
 set(CLP_FFI_PY_LIB_SRC_DIR "${CLP_FFI_PY_SRC_DIR}/clp_ffi_py")
 set(CLP_FFI_PY_CLP_CORE_DIR "${CLP_FFI_PY_SRC_DIR}/clp/components/core")
 
+# Add clp's string utility
```
kirkrodrigues:
```suggestion
# Add CLP's string_utils
```

tests/test_ir/test_json/elasticsearch.jsonl
kirkrodrigues, line 1: Why aren't these in the `test_data` dir?

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -0,0 +1,341 @@
...
+            [[maybe_unused]] clp::ffi::SchemaTree::NodeLocator schema_tree_node_locator
+    ) -> IRErrorCode {
+        return IRErrorCode::IRErrorCode_Success;
+    }
+
+    [[nodiscard]] static auto handle_end_of_stream() -> IRErrorCode {
+        return IRErrorCode::IRErrorCode_Success;
+    }
+
+    // TODO: we should enable linting when clang-tidy config is up-to-date to allow simple classes.
```
kirkrodrigues:
```suggestion
    // TODO: We should enable linting when clang-tidy config is up-to-date to allow simple classes.
```

This is a nit, but try to consistently use capitals after colons.

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.hpp
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * A PyObject structure functioning as a Python-compatible interface to retrieve a key-value pair
+ * log event. The underlying data is pointed to by `m_kv_pair_log_event`.
+ */
+class PyKeyValuePairLogEvent {
+public:
+    // Delete default constructor to disable direct instantiation.
+    PyKeyValuePairLogEvent() = delete;
+
+    // Delete copy/move constructors and assignment operators
```
kirkrodrigues:
```suggestion
    // Delete copy & move constructors and assignment operators
```

Copied from our internal guidelines.

tests/test_ir/test_utils.py
```diff
@@ -22,6 +24,33 @@
 from clp_ffi_py.wildcard_query import WildcardQuery
 
 
+class JsonLinesFileReader:
+    """
+    Class for reading JSON files.
+
+    It assumes each line in the file is a JSON string, and the parser parses each line into a JSON
+    object, and returns then through a generator.
```
kirkrodrigues:
```suggestion
    object, and returns them through a generator.
```

kirkrodrigues (changes requested):

tests/test_ir/test_json/elasticsearch.jsonl
LinZhihao-723: This dir contains raw CLP IR, compressed CLP IR, and some random txt for testing the deserialization buffer.
Creating a new dir for JSON files so we don't have to determine whether a file should be tested or not when walking through the directory. I could move them to `test_data` if you want, but IMO the proper way of doing this is to structure it as:
- test_data
  - test_json
  - test_deserializer_buffer

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
LinZhihao-723: Applied

CMakeLists.txt
```diff
@@ -57,22 +59,37 @@ set(CLP_FFI_PY_SRC_DIR "${CMAKE_CURRENT_SOURCE_DIR}/src")
...
 set(CLP_FFI_PY_CLP_CORE_SOURCES
     ${CLP_FFI_PY_CLP_CORE_DIR}/src/clp/BufferReader.cpp
     ${CLP_FFI_PY_CLP_CORE_DIR}/src/clp/ffi/ir_stream/decoding_methods.cpp
     ${CLP_FFI_PY_CLP_CORE_DIR}/src/clp/ffi/ir_stream/encoding_methods.cpp
+    ${CLP_FFI_PY_CLP_CORE_DIR}/src/clp/ffi/ir_stream/ir_unit_deserialization_methods.cpp
+    ${CLP_FFI_PY_CLP_CORE_DIR}/src/clp/ffi/ir_stream/Serializer.cpp
     ${CLP_FFI_PY_CLP_CORE_DIR}/src/clp/ffi/ir_stream/utils.cpp
     ${CLP_FFI_PY_CLP_CORE_DIR}/src/clp/ffi/encoding_methods.cpp
+    ${CLP_FFI_PY_CLP_CORE_DIR}/src/clp/ffi/SchemaTree.cpp
+    ${CLP_FFI_PY_CLP_CORE_DIR}/src/clp/ffi/KeyValuePairLogEvent.cpp
```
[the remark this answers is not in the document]
LinZhihao-723: The correct file paths are all correct.

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -0,0 +1,343 @@
...
+    PyObjectPtr<PyObject> parsed_json{py_utils_parse_json_str(json_str)};
+    if (nullptr == parsed_json) {
+        return nullptr;
+    }
+    if (false == static_cast<bool>(PyDict_Check(parsed_json.get()))) {
+        PyErr_SetString(PyExc_TypeError, "Serialized JSON object is not a dictionary");
+        return nullptr;
+    }
+    return parsed_json.release();
+}
```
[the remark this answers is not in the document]
LinZhihao-723: No. The implementation is already finished but will be deferred to future PRs

tests/test_ir/test_json/elasticsearch.jsonl
kirkrodrigues: Sure, let's use the structure you proposed.

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -0,0 +1,343 @@
...
+        " However, direct instantiation using the `__init__` method is also supported for testing"
+        " purposes, although this may not be as efficient as emission from the IR deserializer.\n\n"
+        "__init__(self, dictionary)\n\n"
+        "Initializes a :class:`KeyValuePairLogEvent` from the given Python dictionary. Note that"
+        " each object should only be initialized once. Double initialization will result in a"
+        " memory leak.\n\n"
+        ":param dictionary: A dictionary representing the key-value pair log event, where all keys"
+        " must be strings, including keys inside any sub-dictionaries.\n"
+        ":type dictionary: dict[str, Any]\n"
+        ":rtype: int\n"
```
kirkrodrigues: `__init__` returns an int?

tests/test_ir/test_json/elasticsearch.jsonl
kirkrodrigues: Sure, let's use the structure you proposed.

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
LinZhihao-723: hmmm, my bad.
In CPython level, `__init__` returns an int to indicate error (when non-zero). But I guess it's just the signature for the callback.
In the actual Python `__init__`, it should return None: https://codeql.github.com/codeql-query-help/python/py-explicit-return-in-init/

LinZhihao-723: Removed

tests/test_ir/test_json/elasticsearch.jsonl
LinZhihao-723: Moved jsonl files into `test_data/jsonl`
I didn't move the rest of files yet to not complicate this PR

tests/test_ir/test_key_value_pair_log_event.py
```diff
@@ -0,0 +1,41 @@
...
+from clp_ffi_py.ir import KeyValuePairLogEvent
+
+
+class TestCaseKeyValuePairLogEvent(TestCLPBase):
+    """
+    Class for testing `clp_ffi_py.ir.KeyValuePairLogEvent`.
+    """
+
+    test_data_dir: Path = Path("test_data")
+    jsonl_dir: Path = Path("jsonl")
```
kirkrodrigues: * I find it a bit odd that these are paths since they don't actually point at anything on disk until they're resolved against `current_dir`. Personally, I would just make them strings.
* If you do make them strings, how about calling them `test_data_dir_name` and `jsonl_test_data_dir_name`?
* If you don't make them strings, how about calling `jsonl_dir` -> `jsonl_test_data_dir`?

tests/test_ir/test_key_value_pair_log_event.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def test_basic(self) -> None:
+        """
+        Tests the conversion between a Python dictionary and a `KeyValuePairLogEvent` instance,
+        ensuring accurate serialization and deserialization in both directions.
+        """
+        current_dir: Path = Path(__file__).resolve().parent
+        test_src_dir: Path = (
```
kirkrodrigues: This should really be called `test_data_dir`, right?

kirkrodrigues (commented):
One last set of suggestions.

tests/test_ir/test_key_value_pair_log_event.py
LinZhihao-723: how about:
```
jsonl_test_data_dir: Path = Path("test_data") / Path("jsonl")
```

tests/test_ir/test_key_value_pair_log_event.py
LinZhihao-723: yeah make sense

kirkrodrigues (approved):

PR #82, opened 2024-11-04
Fix the stub file of the IR native module.

LinZhihao-723:
# References

# Description

The current stub file of `clp_ffi_py.ir.native` has an issue when referring to static methods. This PR adds type object before the static methods.

# Validation performed

- Ensure workflow passed

## Summary by CodeRabbit

- **New Features**
	- Enhanced clarity in the `Query` class constructor by updating default parameter values to explicitly call class methods, ensuring consistency in default sourcing.

### DIFF
--- a/clp_ffi_py/ir/native.pyi
+++ b/clp_ffi_py/ir/native.pyi
@@ -43,10 +43,10 @@ class Query:
     def default_search_time_termination_margin() -> int: ...
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


LinZhihao-723:
Reading more, realized the stub file shouldn't have default values rather than "simple" values: https://docs.astral.sh/ruff/rules/argument-default-in-stub/
We should replace them by `...` instead

PR #83, opened 2024-11-05
feat: Add Python class `Serializer` to serialize log events into CLP key-value pair IR format.

LinZhihao-723:
# References

# Description

This PR depends on #81.
This PR adds a new class `Serializer` to support serializing log events (as msgpack maps) into CLP key-value pair IR format. The `Serializer` is designed to write to a Python `IO[byte]` object as the output stream. It is also designed to be compatible with Python context manager.

# Validation performed

- Ensure all unit tests passed, with the new ones added to test JSON file serialization.
- Ensure workflow passed.
- Ran with address sanitizer to ensure no memory errors
- Tested in the beta branch #86 

## Release Notes

- **New Features**
  - Introduced the `Serializer` class for enhanced log event serialization.
  - Added `KeyValuePairLogEvent` and `Serializer` to the public API.
  
- **Improvements**
  - Dynamic project name and version assignment in the build configuration.
  - Enhanced error handling and resource management in the `Serializer` class.

- **Tests**
  - Added comprehensive tests for the `Serializer` class, including JSON serialization and buffer management scenarios.


tests/test_ir/test_key_value_pair_log_event.py
```diff
@@ -0,0 +1,31 @@
...
+        test_src_dir: Path = current_dir / Path(TestCaseKeyValuePairLogEvent.input_src_dir)
+        for file_path in test_src_dir.rglob("*"):
+            if not file_path.is_file():
+                continue
+            json_file_reader: JsonFileReader = JsonFileReader(file_path)
+            for expected in json_file_reader.read_json_lines():
+                self.assertIsInstance(expected, dict, "Input must be a dictionary")
+                actual: KeyValuePairLogEvent = KeyValuePairLogEvent(expected)
+                serialized_py_dict: Dict[Any, Any] = actual.to_dict()
+                self.assertEqual(expected, serialized_py_dict)
```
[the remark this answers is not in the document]
LinZhihao-723: Not in this PR

tests/test_ir/test_serializer.py
```diff
@@ -38,3 +43,123 @@
...
+                serializer.flush()
+                self.assertEqual(len(byte_buffer.getvalue()), serializer.get_num_bytes_serialized())
+                self.assertEqual(num_bytes_serialized, serializer.get_num_bytes_serialized())
+            serializer.close()
+            self.assertEqual(
+                num_bytes_serialized + 1,
+                serializer.get_num_bytes_serialized(),
+                "End-of-stream byte is missing",
+            )
+
```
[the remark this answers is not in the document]
LinZhihao-723: The point is to test `with`... so there might be some other ways to de-duplicate the code. Will come back and check later

clp_ffi_py/ir/native.pyi
```diff
@@ -82,4 +85,22 @@ class FourByteDeserializer:
         allow_incomplete_stream: bool = False,
     ) -> Optional[LogEvent]: ...
 
+class KeyValuePairLogEvent:
+    def __init__(self, dictionary: Dict[Any, Any]): ...
+    def to_dict(self) -> Dict[Any, Any]: ...
```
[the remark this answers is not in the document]
LinZhihao-723: Not in this PR

src/clp_ffi_py/Py_utils.cpp
```diff
@@ -77,12 +87,36 @@ auto py_utils_get_timezone_from_timezone_id(std::string const& timezone_id) -> P
...
+}
+
+auto py_utils_parse_json_str(std::string_view json_str) -> PyObject* {
+    PyObjectPtr<PyObject> const func_args_ptr{Py_BuildValue("(s)", json_str.data())};
     auto* func_args{func_args_ptr.get()};
     if (nullptr == func_args) {
         return nullptr;
     }
-    return py_utils_function_call_wrapper(Py_func_serialize_dict_to_msgpack.get(), func_args);
+    return py_utils_function_call_wrapper(Py_func_parse_json_str.get(), func_args);
```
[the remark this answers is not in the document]
LinZhihao-723: Python will throw exceptions in that case.

tests/test_ir/test_utils.py
```diff
@@ -22,6 +24,32 @@
...
+        self.file_path: Path = file_path
+
+    def read_json_lines(self) -> Generator[Any, None, None]:
+        """
+        Reads each line in the JSON file, parses it as a JSON object, and yields the JSON object.
+
+        :yield: Parsed JSON object for each line in the file.
+        """
+        with open(self.file_path, "r", encoding="utf-8") as file:
+            yield from (json.loads(line.strip()) for line in file)
```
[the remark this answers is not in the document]
LinZhihao-723: Not in this PR

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.hpp
```diff
@@ -0,0 +1,76 @@
...
+     * Initializes the underlying data with the given inputs. Since the memory allocation of
+     * `PyKeyValuePairLogEvent` is handled by CPython's allocator, cpp constructors will not be
+     * explicitly called. This function serves as the default constructor initialize the underlying
+     * key-value pair log event. It has to be manually called whenever creating a new
+     * `PyKeyValuePairLogEvent` object through CPython APIs.
+     * @param kv_pair_log_event
+     * @return true on success.
+     * @return false on failure with the relevant Python exception and error set.
+     */
+    [[nodiscard]] auto init(clp::ffi::KeyValuePairLogEvent kv_pair_log_event) -> bool;
```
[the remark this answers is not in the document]
LinZhihao-723: Not in this PR

src/clp_ffi_py/ir/native/PySerializer.hpp
```diff
@@ -0,0 +1,179 @@
...
+     * It has to be manually called whenever creating a new `PySerializer` object through
+     * CPython APIs.
+     * @param output_stream
+     * @param serializer
+     * @param buffer_size_limit
+     * @return true on success.
+     * @return false on failure with the relevant Python exception and error set.
+     */
+    [[nodiscard]] auto
+    init(PyObject* output_stream, ClpIrSerializer serializer, Py_ssize_t buffer_size_limit) -> bool;
```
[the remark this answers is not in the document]
LinZhihao-723: we use `std::move` to construct

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -0,0 +1,338 @@
...
+        return std::nullopt;
+    }
+
+    return std::move(ir_unit_handler.m_log_event);
+}
+
+CLP_FFI_PY_METHOD auto PyKeyValuePairLogEvent_dealloc(PyKeyValuePairLogEvent* self) -> void {
+    self->clean();
+    PyObject_Del(self);
+}
```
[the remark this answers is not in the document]
LinZhihao-723: Applied.

src/clp_ffi_py/ir/native/PySerializer.hpp
```diff
@@ -0,0 +1,191 @@
...
+        m_buffer_size_limit = 0;
+    }
+
+    /**
+     * Releases the memory allocated for underlying data fields.
+     */
+    auto clean() -> void {
+        close_serializer();
+        Py_XDECREF(m_output_stream);
+    }
```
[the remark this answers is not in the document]
LinZhihao-723: It never fails.

LinZhihao-723 (commented):
The API and API doc strings have been reviewed in #86 by @kirkrodrigues

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.hpp
```diff
@@ -0,0 +1,88 @@
...
+    PyKeyValuePairLogEvent() = delete;
+
+    // Delete copy/move constructors and assignments
+    PyKeyValuePairLogEvent(PyKeyValuePairLogEvent const&) = delete;
+    PyKeyValuePairLogEvent(PyKeyValuePairLogEvent&&) = delete;
+    auto operator=(PyKeyValuePairLogEvent const&) -> PyKeyValuePairLogEvent& = delete;
+    auto operator=(PyKeyValuePairLogEvent&&) -> PyKeyValuePairLogEvent& = delete;
+
+    // Destructor
+    ~PyKeyValuePairLogEvent() = default;
```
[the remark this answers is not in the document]
LinZhihao-723: We use `clean()` for destruction

src/clp_ffi_py/PyObjectUtils.hpp
```diff
@@ -47,6 +49,32 @@ using PyObjectPtr = std::unique_ptr<PyObjectType, PyObjectDeleter<PyObjectType>>
...
+ * A guard class for Python exceptions. In certain CPython methods, such as `tp_finalize`,
+ * the exception state must remain unchanged throughout execution. This class saves the current
+ * exception state upon initialization and restores it upon destruction, ensuring the exception
+ * status is preserved.
+ * Docs: https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_finalize
+ */
+class PyErrGuard {
+public:
+    // Constructor
+    PyErrGuard() = default;
```
haiqi96, line 63: Very very Nit: I understand that we are relying on m_exception_context to be default initialized to "capture" the exception state. But this is still quite unintuitive for me. 

Wonder is there any way to make it more obvious that m_exception_context is initialized to "capture" the exception state in the constructor? It's fine it we keep this current version as well.

tests/test_ir/test_serializer.py
```diff
@@ -38,3 +43,127 @@ def test_serialization_methods_consistency(self) -> None:
...
+        The JSON parser will parse the file into Python dictionaries,
+        and then convert them into msgpack and feed into `clp_ffi_py.ir.Serializer`.
+        """
+
+        test_files: List[Path] = self.__get_test_files()
+        byte_buffer: BytesIO
+        num_bytes_serialized: int
+        serializer: Serializer
+
+        # Test with context manager
```
haiqi96, line 70: These are very simple test cases. I assume that's because you need deserializer for more thorough testing?

src/clp_ffi_py/ir/native/PySerializer.hpp
```diff
@@ -0,0 +1,190 @@
...
+    PySerializer() = delete;
+
+    // Delete copy & move constructors and assignment operators
+    PySerializer(PySerializer const&) = delete;
+    PySerializer(PySerializer&&) = delete;
+    auto operator=(PySerializer const&) -> PySerializer& = delete;
+    auto operator=(PySerializer&&) -> PySerializer& = delete;
+
+    // Destructor
+    ~PySerializer() = default;
```
haiqi96, line 37: Does Default destructor delete m_output_stream?

src/clp_ffi_py/ir/native/PySerializer.hpp
```diff
@@ -0,0 +1,190 @@
...
+        Py_XDECREF(m_output_stream);
+    }
+
+    [[nodiscard]] auto is_closed() const -> bool { return nullptr == m_serializer; }
+
+    /**
+     * Asserts the serializer has not been closed.
+     * @return true on success, false if it's already been closed with `IOError` set.
+     */
+    [[nodiscard]] auto assert_is_not_closed() const -> bool;
```
haiqi96: Can this be a private member function? Don't see it gets used anywhere in this PR

src/clp_ffi_py/ir/native/PySerializer.hpp
```diff
@@ -0,0 +1,190 @@
...
+
+    /**
+     * Serializes the log event from the given msgpack map into IR format.
+     * NOTE: the serializer must not be closed to call this method.
```
haiqi96: Nit: I don't think we need this note? It should be obvious and your function is also properly handling the unopened case.

The note made me feel your implementation is not checking if the seralizer is closed. It doesn't hurt to keep the comment though

src/clp_ffi_py/ir/native/PySerializer.hpp
```diff
@@ -0,0 +1,190 @@
...
+     * @return true on success.
+     * @return false on failure with the relevant Python exception and error set.
+     */
+    [[nodiscard]] auto flush() -> bool;
+
+    /**
+     * Closes the serializer by writing the buffered results into the output stream with
+     * end-of-stream IR Unit appended in the end.
+     * NOTE: the serializer must not be closed to call this method.
```
haiqi96: ditto

src/clp_ffi_py/ir/native/PySerializer.cpp
```diff
@@ -0,0 +1,576 @@
...
+namespace clp_ffi_py::ir::native {
+namespace {
+/**
+ * Callback of `PySerializer`'s `__init__` method:
+ */
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays)
+PyDoc_STRVAR(
+        cPySerializerDoc,
+        "Serializer for serializing CLP key-value pair IR streams.\n"
+        "This class serializes log events using the CLP key-value pair IR format and writes the"
```
haiqi96: how about
```suggestion
        "This class serializes log events into the CLP key-value pair IR format and writes the"
```

src/clp_ffi_py/ir/native/PySerializer.cpp
```diff
@@ -0,0 +1,576 @@
...
+        " serialized data to a specified byte stream object.\n\n"
+        "__init__(self, output_stream, buffer_size_limit=65536)\n\n"
+        "Initializes a :class:`Serializer` instance with the given output stream. Note that each"
+        " object should only be initialized once. Double initialization will result in a memory"
+        " leak.\n\n"
+        ":param output_stream: A writable byte output stream to which the serializer will write the"
+        " serialized IR byte sequences.\n"
+        ":type output_stream: IO[bytes]\n"
+        ":param buffer_size_limit: Threshold of how much serialized data to buffer before flushing"
```
haiqi96:
```suggestion
        ":param buffer_size_limit: The maximum amount of serialized data to buffer before flushing"
```

haiqi96 (commented):
first batch of reviews

src/clp_ffi_py/ir/native/PySerializer.cpp
```diff
@@ -0,0 +1,576 @@
...
+                &output_stream,
+                &buffer_size_limit
+        )))
+    {
+        return -1;
+    }
+
+    // Ensure the `output_stream` has `write`, `flush`, and `close` methods
+    auto output_stream_has_method = [&](char const* method_name) -> bool {
+        PyObjectPtr<PyObject> const write_method{PyObject_GetAttrString(output_stream, method_name)
```
haiqi96: Can it be simply method instead of write_method?

src/clp_ffi_py/ir/native/PySerializer.cpp
```diff
@@ -0,0 +1,576 @@
...
+        return -1;
+    }
+    if (false == output_stream_has_method("flush")) {
+        return -1;
+    }
+    if (false == output_stream_has_method("close")) {
+        return -1;
+    }
+
+    if (0 > buffer_size_limit) {
```
haiqi96, line 265: do we allow buffer_size_limit = 0?

src/clp_ffi_py/ir/native/PySerializer.cpp
```diff
@@ -0,0 +1,576 @@
...
+CLP_FFI_PY_METHOD auto PySerializer_dealloc(PySerializer* self) -> void {
+    PyErrGuard const err_guard;
+
+    if (false == self->is_closed()) {
+        if (0
+            != PyErr_WarnEx(
+                    PyExc_ResourceWarning,
+                    "`Serializer.close()` is not called before object destruction. Forget to call "
+                    "`close` will leave the stream incomplete, and potentially resulting in data "
+                    "loss due to data buffering",
```
haiqi96: Nit:
and I believe we should keep the space at the beginning of the line
```suggestion
                    "`Serializer.close()` is not called before object destruction, which will leave"
                    " the stream incomplete, and potentially resulting in data"
                    " loss due to data buffering",
```

haiqi96 (commented):
Another small batch

src/clp_ffi_py/PyObjectUtils.hpp
LinZhihao-723: hmmmm, I think the point is we can rely on the docstring to help readers understand what this class is used for, instead of trying to read the implementation and figure out what's going on.
That said, I don't think we should try to polish the implementation side. Instead, do u think we should improve any documentation of this class?

haiqi96: I guess it's ok, it doesn't take very long for me to realize what it is doing.

src/clp_ffi_py/ir/native/PySerializer.cpp
LinZhihao-723: uh yeah it's a mistake, good catch

src/clp_ffi_py/ir/native/PySerializer.cpp
LinZhihao-723: That's supposed to mean writing to the output stream for every write

src/clp_ffi_py/PyObjectUtils.hpp
LinZhihao-723: kk, then I will say probably just leave it as it is

src/clp_ffi_py/ir/native/PySerializer.cpp
LinZhihao-723: Yeah make sense, mb

src/clp_ffi_py/ir/native/PySerializer.hpp
LinZhihao-723: No, default destructor will not call delete on any pointer typed members.
We have a static check to ensure the object is trivially destructable

tests/test_ir/test_serializer.py
LinZhihao-723: yeah, these test cases are basically to cover functionalities such as `flush` and context manager.
We will have serder tests in the deserializer PR

src/clp_ffi_py/ir/native/PySerializer.cpp
haiqi96: Had an offline discussion. It seems that when buffer_size_limit = 0, there will still be a step of buffering. Any write will be buffered but then immediately flushed to output stream before next write.

This is a little counter-intuitive as buffer_size_limit = 0 make me thought there won't be an internal data.

That said, from observed behavior, user will see the data gets flushed out immediately, so buffer_size_limit = 0 makes sense.

Ideally it would be great if we can get rid of the internal buffering, but looking at the code, this could be really hard. so maybe we live with this for now. Maybe we can document this behavior somewhere? doesn't need to be in Docstring imo.

src/clp_ffi_py/ir/native/PySerializer.hpp
haiqi96: to make sure I understand, does the static check analyze the code and guarantee that the pointer member won't be pointing to memory dynamically allocated by the class?

src/clp_ffi_py/ir/native/PySerializer.hpp
```diff
@@ -0,0 +1,190 @@
...
+        m_serializer = nullptr;
+    }
+
+    /**
+     * Wrapper of `output_stream`'s `write` method.
+     * @param buf
+     * @return The number of bytes written on success.
+     * @return std::nullopt on failure with the relevant Python exception and error set.
+     */
+    [[nodiscard]] auto write_to_output_stream(BufferView buf) -> std::optional<Py_ssize_t>;
```
haiqi96, line 159: why we use optional as return type instead of result?

src/clp_ffi_py/ir/native/PySerializer.hpp
LinZhihao-723: This is the static check we use:
```
static_assert(std::is_trivially_destructible<PySerializer>());
```
https://en.cppreference.com/w/cpp/language/destructor#Trivial_destructor

And I don't think any non-user provided constructor will call `delete`. If `PySerializer` is trivially destructible, then there should be no `delete` called in any of the constructors in any of the non-static members

src/clp_ffi_py/ir/native/PySerializer.cpp
LinZhihao-723: The reason to use internal buffering is to avoid calling Python level `IO[bytes].write` every time (which has overhead). This is also the behavior of normal compressor, right? Like zstd provides a configurable buffer size, so I think exposing such an API should be good. We can polish the doc string to make it more clear

src/clp_ffi_py/ir/native/PySerializer.hpp
LinZhihao-723: Sure, let me remove them

src/clp_ffi_py/ir/native/PySerializer.hpp
LinZhihao-723: are u talking about `boost::std_result`?

src/clp_ffi_py/ir/native/PySerializer.cpp
LinZhihao-723: > Any write will be buffered but then immediately flushed to output stream before next write.`

btw, this is not correct. When the buffer_size_limit is 0, the buffered data will be flushed to the output stream immediately after serialization returns success, not before the next write. So technically speaking, the buffer is used as an intermediate storage, and in users' view, there's no internal buffering.

src/clp_ffi_py/ir/native/PySerializer.cpp
```diff
@@ -0,0 +1,575 @@
...
+    Py_ssize_t buffer_size_limit{PySerializer::cDefaultBufferSizeLimit};
+    if (false
+        == static_cast<bool>(PyArg_ParseTupleAndKeywords(
+                args,
+                keywords,
+                "O|L",
+                static_cast<char**>(keyword_table),
+                &output_stream,
+                &buffer_size_limit
+        )))
```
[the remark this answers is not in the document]
LinZhihao-723: That's a good catch. Changed.

src/clp_ffi_py/ir/native/PySerializer.hpp
haiqi96: yes, wonder is there a reason or is it just hard to include boost

LinZhihao-723: we just don't need it. When using `std_result`, it is assumed the caller will need to know the details of the error. But in CPython, the error is set by the callee through CPython's exception APIs. The caller only needs to know whether an error happened, but doesn't need to know any details about the error.

haiqi96 (approved):
The title looks good to me

PR #84, opened 2024-11-06
Add required error handling steps in `PyDeserializerBuffer::py_getbuffer`.

LinZhihao-723:
# References

# Description

When running unit tests in a Python debug build with address sanitizer enabled, we get the following error:
```
test_buffer_protocol (test_ir.test_deserializer_buffer.TestCaseDeserializerBuffer)
Tests whether the buffer protocol is disabled by default. ... Fatal Python error: _Py_CheckSlotResult: Slot getbuffer of type clp_ffi_py.ir.native.DeserializerBuffer failed without setting an exception
Python runtime state: initialized
```
This is because we return failure in `PyDeserializerBuffer_getbuffer` but never set an exception. It was a mistake in the current implementation and this PR fixes the issue.

# Validation performed

- Ensure workflow passed.
- Ensure all unit tests passed with Python (debug build with address sanitizer)

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling for various methods in the `PyDeserializerBuffer`, providing clearer error messages.

- **Bug Fixes**
	- Improved robustness of error reporting for buffer access and consumption issues.

### DIFF
--- a/src/clp_ffi_py/ir/native/PyDeserializerBuffer.cpp
+++ b/src/clp_ffi_py/ir/native/PyDeserializerBuffer.cpp
@@ -302,7 +302,6 @@ auto PyDeserializerBuffer::py_getbuffer(Py_buffer* view, int flags) -> int {
     if (false == is_py_buffer_protocol_enabled()) {
         // The steps below are required by the spec
         // https://docs.python.org/3/c-api/typeobj.html#c.PyBufferProcs.bf_getbuffer
-        // https://docs.python.org/3/c-api/typeobj.html#c.PyBufferProcs.bf_getbuffer
         view->obj = nullptr;
         PyErr_SetString(
                 PyExc_BufferError,
--- a/src/clp_ffi_py/ir/native/PyDeserializerBuffer.cpp
+++ b/src/clp_ffi_py/ir/native/PyDeserializerBuffer.cpp
@@ -300,8 +300,11 @@ auto PyDeserializerBuffer::metadata_init(PyMetadata* metadata) -> bool {
 
 auto PyDeserializerBuffer::py_getbuffer(Py_buffer* view, int flags) -> int {
     if (false == is_py_buffer_protocol_enabled()) {
+        // `view->obj` should be set to NULL according to the doc:
+        // https://docs.python.org/3/c-api/typeobj.html#c.PyBufferProcs.bf_getbuffer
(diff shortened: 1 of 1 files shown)


src/clp_ffi_py/ir/native/PyDeserializerBuffer.cpp
```diff
@@ -299,9 +299,12 @@ auto PyDeserializerBuffer::metadata_init(PyMetadata* metadata) -> bool {
 }
 
 auto PyDeserializerBuffer::py_getbuffer(Py_buffer* view, int flags) -> int {
-    // Don't need to set the exception message during the failure. The Python level caller will set
-    // the exception and thus overwrite it.
     if (false == is_py_buffer_protocol_enabled()) {
+        PyErr_SetString(
```
kirkrodrigues, line 306: So is the implication of this PR that we should *always* set the exception even if the caller will overwrite it?

LinZhihao-723: No, the comment was wrong. We should raise an exception according to the doc:
https://docs.python.org/3/c-api/typeobj.html#c.PyBufferProcs.bf_getbuffer

(I also fixed the implementation by raising `BufferError` and set `view->obj` to NULL

src/clp_ffi_py/ir/native/PyDeserializerBuffer.cpp
```diff
@@ -299,9 +299,15 @@ auto PyDeserializerBuffer::metadata_init(PyMetadata* metadata) -> bool {
 }
 
 auto PyDeserializerBuffer::py_getbuffer(Py_buffer* view, int flags) -> int {
-    // Don't need to set the exception message during the failure. The Python level caller will set
-    // the exception and thus overwrite it.
     if (false == is_py_buffer_protocol_enabled()) {
+        // `view->obj` should be set to NULL according to the doc:
```
kirkrodrigues:
```suggestion
        // The steps below are required by the spec:
        // https://docs.python.org/3/c-api/typeobj.html#c.PyBufferProcs.bf_getbuffer
```

kirkrodrigues (approved):
For the PR title, how about:

```
Add required error handling steps in `PyDeserializerBuffer::py_getbuffer`.
```

PR #86, opened 2024-11-10
feat: Add support for key-value pair IR format serialization and deserialization.

LinZhihao-723:
# References

# Description

This PR adds support for key-value pair IR format serialization and deserialization:
- `Serializer` for IR serialization
- `Deserializer` for IR deserialization
- `KeyValuePairLogEvent` for storing deserialized log events

# Validation performed

- Add unit tests to cover newly introduced classes
- Ensure workflow builds successfully
- Run `Debug` build in Python 3.10 debug build + address sanitizer, ensure no errors/warnings caused by our implementation

PR #87, opened 2024-11-11
feat: Add `PyExceptionContext` for passing Python exceptions through C++ exceptions.

LinZhihao-723:
# Description

Before this PR, we could not pass Python exceptions around when throwing/catching C++ exceptions.
This PR introduces class `PyExceptionContext` to solve this problem, allowing users to store/restore Python exceptions. Python exceptions can be fetched/restored through `PyErr_Fetch` and `PyErr_Restore` as explained [here](https://docs.python.org/3/c-api/exceptions.html#c.PyErr_Fetch). `PyExceptionContext` is essentially a wrapper to these two methods.
`PyExceptionContext` is then used to improve python ffi's implementation of `clp::TraceableException`, allowing use to pass Python exceptions around through the traceable exception. This is required in some other clp interface implementations, for example, if we want to implement reader/writer interfaces that directly communicate with Python objects, we need a way to caught the Python exceptions through exception throwing. Examples can be found in #86's `DeserializerBufferReader`.
This PR also adds a general exception handling helper to convert a C++ traceable exception into Python exceptions. This method is used to replace the existing exception handling implementations.

# Validation performed

- Ensure the workflow passed
- Ensure all clang-tidy warnings are resolved in the newly added code sections except non-directly included Python headers
- Tested within the beta branch

## Summary by CodeRabbit

- **New Features**
	- Introduced a new class for enhanced management of Python exceptions.
	- Added a method to access the Python exception context within the exception handling class.
	- Implemented a new function for structured error handling of traceable exceptions.

- **Bug Fixes**
	- Updated exception handling in relevant classes to utilize a more specific exception type for better traceability.

- **Documentation**
	- Enhanced documentation for the exception handling class to clarify its capabilities.

### DIFF
--- a/src/clp_ffi_py/PyExceptionContext.hpp
+++ b/src/clp_ffi_py/PyExceptionContext.hpp
@@ -1,0 +1,68 @@
+#ifndef CLP_FFI_PY_PYEXCEPTIONCONTEXT_HPP
+#define CLP_FFI_PY_PYEXCEPTIONCONTEXT_HPP
+
+#include <clp_ffi_py/Python.hpp>  // Must always be included before any other header files
+
+namespace clp_ffi_py {
+/**
+ * Class to get/set Python exception context, designed to capture the current exception state upon
+ * instantiation and provide a restore API that allows users to reinstate the exception context when
+ * needed. Doc: https://docs.python.org/3/c-api/exceptions.html#c.PyErr_Fetch
+ */
+class PyExceptionContext {
+public:
+    // Constructor
+    /**
+     * Constructs the context by fetching the current raised exception (if any).
+     */
+    PyExceptionContext() { PyErr_Fetch(&m_type, &m_value, &m_traceback); }
+
+    // Delete copy/move constructors and assignments
+    PyExceptionContext(PyExceptionContext const&) = delete;
+    PyExceptionContext(PyExceptionContext&&) = delete;
+    auto operator=(PyExceptionContext const&) -> PyExceptionContext& = delete;
+    auto operator=(PyExceptionContext&&) -> PyExceptionContext& = delete;
+
+    ~PyExceptionContext() {
+        Py_XDECREF(m_type);
+        Py_XDECREF(m_value);
+        Py_XDECREF(m_traceback);
+    }
+
+    /**
+     * @return Whether the context stores an exception.
+     */
+    [[nodiscard]] auto has_exception() const noexcept -> bool { return nullptr != m_value; }
+
+    /**
+     * Restores the exception by the context.
+     * NOTE:
+     * - This method will clear the existing exception if one is set.
+     * - The stored context will be cleared after restoration.
+     * - If there's no exception in the stored context, the error indicator will be cleared.
+     * - This method should be called strictly once, otherwise the error indicator will be cleared.
+     * @return Whether an exception has been set by restoring the context.
+     */
+    [[maybe_unused]] auto restore() noexcept -> bool {
+        auto const exception_has_been_set{has_exception()};
+        PyErr_Restore(m_type, m_value, m_traceback);
+        m_type = nullptr;
+        m_value = nullptr;
+        m_traceback = nullptr;
+        return exception_has_been_set;
+    }
+
+    [[nodiscard]] auto get_type() const -> PyObject* { return m_type; }
+
+    [[nodiscard]] auto get_value() const -> PyObject* { return m_value; }
+
+    [[nodiscard]] auto get_traceback() const -> PyObject* { return m_traceback; }
+
+private:
+    PyObject* m_type{nullptr};
+    PyObject* m_value{nullptr};
+    PyObject* m_traceback{nullptr};
+};
+}  // namespace clp_ffi_py
+
+#endif  // CLP_FFI_PY_PYEXCEPTIONCONTEXT_HPP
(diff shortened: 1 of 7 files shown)


LinZhihao-723:
@coderabbitai review

src/clp_ffi_py/PyExceptionContext.hpp
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    // Methods
+    /**
+     * @return Whether the context stores an exception.
+     */
+    [[nodiscard]] auto has_exception() const noexcept -> bool { return nullptr != m_value; }
+
+    /**
+     * Restores the exception by the context.
```
haiqi96: nit: from the context?

src/clp_ffi_py/utils.cpp
```diff
@@ -59,4 +62,22 @@ auto get_py_bool(bool is_true) -> PyObject* {
     }
     Py_RETURN_FALSE;
 }
+
+auto handle_traceable_exception(clp::TraceableException& exception) noexcept -> void {
+    if (auto* py_ffi_exception{dynamic_cast<ExceptionFFI*>(&exception)}) {
+        auto& exception_context{py_ffi_exception->get_py_exception_context()};
+        if (exception_context.has_exception() && exception_context.restore()) {
```
haiqi96: why we are &&ing `exception_context.has_exception()` and `exception_context.restore()`? Is it possible that the 2nd expression evaluates to false while the first is true?

it looks like exception_context.restore() will not have any effect if it doesn't have an exception, so wonder do we really need `has_exception()` method?

LinZhihao-723: 1. In the current implementation, if `has_exception` returns true, `restore` will always return true. However, I think it might be better to keep track of `restore`'s return value to ensure that a Python exception has been set before returning.
2. If `has_exception` returns false, `restore` will clear any exceptions already set

haiqi96: >If has_exception returns false, restore will clear any exceptions already set

Is this behavior desired in some cases? At least in this line, I can see that you are preventing this from happening? (i.e., you are calling has_exception() first and will not proceed to call restore() if has_exception is false, so that the exception that is already set won;t be cleared

LinZhihao-723: 1. The point of adding this check is we don't want to restore an empty exception
2. Yes, it will be used if we want to store exception states and recover it later

haiqi96: what if we write the restore in this way, so technically you don't need to call an extra `has_exception()`.
```
[[no_discard]] auto restore() noexcept -> bool {
    auto const exception_has_been_set{has_exception()};
    if (exception_has_been_set) {
      PyErr_Restore(m_type, m_value, m_traceback);
      m_type = nullptr;
      m_value = nullptr;
      m_traceback = nullptr;
    }
    return exception_has_been_set;
}
```

My point is that, it's quite redundant to require user to call `has_exception` if you still need to check the return value of restore().
```if (exception_context.has_exception() && exception_context.restore()) {```
In the line above, do you see a case where checking the return value of `restore()` is important and can avoid corner case, when ``has_exception()``` already returns true.

Comment (LinZhihao-723) [reply src/clp_ffi_py/utils.cpp]: This will prevent us from restoring the empty exception.
Check this example: https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_finalize
We may want to recover the exception state before returning, meaning that we may need to restore an empty exception

Comment (LinZhihao-723) [reply src/clp_ffi_py/utils.cpp]: Discussed offline and we decide to move `exception_context.restore()` into the if statement

File: src/clp_ffi_py/ir/native/PyMetadata.cpp [hunk 3]
```diff
@@ -237,13 +239,8 @@ auto PyMetadata::init(nlohmann::json const& metadata, bool is_four_byte_encoding
...
         // NOLINTNEXTLINE(cppcoreguidelines-owning-memory)
         m_metadata = new Metadata(metadata, is_four_byte_encoding);
-    } catch (ExceptionFFI const& ex) {
-        PyErr_Format(
-                PyExc_RuntimeError,
-                "Failed to initialize metadata from deserialized JSON format preamble. "
-                "Error message: %s",
-                ex.what()
-        );
+    } catch (clp::TraceableException& ex) {
```
haiqi96: One last comment:

By replacing ExceptionFFI with clp::TraceableException, now we need to use dynamic cast in the handle_traceable_exception.

Just by looking at this PR, I don't see the benefit of replacing ExceptionFFI with a more general clp::TraceableException, but I guess you have some insights/future plan?

LinZhihao-723: hmmm, for this one I think it is fine to keep using `clp_ffi_py::ExceptionFFI`, and maybe we should since it's the only type of exception it could throw. In future PRs, we may also need to capture exceptions from clp core code execution, so it's better to catch a generic type instead.
Changed it back to `clp_ffi_py::ExceptionFFI` in the latest push

haiqi96 (approved):
How about:
Add class PyExceptionContext for handling Python exceptions in C++.

LinZhihao-723:
> How about: Add class PyExceptionContext for handling Python exceptions in C++.

The class is not for handling Python exceptions in C++ right? Like it is more like a wrapper to pass exceptions around. This PR improves the error handling but not building handling semantics from scratch

LinZhihao-723:
Discussed offline and agreed on the final PR title.

PR #89, opened 2024-11-16
feat: Add `DeserializerBufferReader` as a `clp::ReaderInterface` implementation to directly read from a Python `IO[bytes]` object.

LinZhihao-723:
# Description

In the current code base, we have implemented `DeserializerBuffer` as a Python object with a buffer protocol that is supported to directly read data from a Python `IO[bytes]` object. However, such an object doesn't have any `clp::ReaderInterface` support, making it hard to use any IR deserialization method directly. We used `clp::BufferReader` as a workaround to wrap the underlying data buffer and provide reading capabilities. However, this means we need to add additional logic to handle the case where the buffer is truncated and manually try to read more when needed. From our previous experience, this solution is painful, error-prone, and has potential performance issues.
As an improvement, this PR introduces a new class, `DeserializerBufferReader`, to implement `clp::ReaderInterface` as a wrapper of `DeserialzierBuffer`. It uses `DeserializerBuffer` as an underlying data buffer of bytes read from the Python level `IO[bytes]` stream, and it automatically handles any read requests by implementing `try_read`. This makes the deserialization methods easier to implement and more readable/maintainable.

# Validation performed

- Ensure workflow passed
- Tested in the beta branch to implement deserializer by reading IR stream with different buffer capacities. Workflow passed with relevant unit tests all passed: #86 

## Summary by CodeRabbit

- **New Features**
	- Introduced the `DeserializerBufferReader` class for enhanced data reading from Python byte streams.
	- Added factory methods for creating instances of `DeserializerBufferReader` and `PyDeserializerBuffer`.

- **Bug Fixes**
	- Improved error handling in buffer reading operations.

- **Documentation**
	- Updated header files to include necessary directives for better integration.

- **Chores**
	- Enhanced build configuration for improved project management and library installation.

### DIFF
--- a/src/clp_ffi_py/ir/native/DeserializerBufferReader.cpp
+++ b/src/clp_ffi_py/ir/native/DeserializerBufferReader.cpp
@@ -56,7 +56,7 @@ auto DeserializerBufferReader::try_read(char* buf, size_t num_bytes_to_read, siz
             );
         }
         num_bytes_read += num_bytes_copied;
-        dst_buf = dst_buf.last(dst_buf.size() - num_bytes_copied);
+        dst_buf = dst_buf.subspan(num_bytes_copied);
     }
 
     return clp::ErrorCode_Success;
--- a/src/clp_ffi_py/ir/native/DeserializerBufferReader.cpp
+++ b/src/clp_ffi_py/ir/native/DeserializerBufferReader.cpp
@@ -1,0 +1,83 @@
+#include <clp_ffi_py/Python.hpp>  // Must always be included before any other header files
+
+#include "DeserializerBufferReader.hpp"
+
+#include <algorithm>
+#include <cstddef>
+#include <cstdint>
+#include <span>
+
+#include <clp/ErrorCode.hpp>
+#include <clp/type_utils.hpp>
+#include <gsl/gsl>
+
+#include <clp_ffi_py/ExceptionFFI.hpp>
+#include <clp_ffi_py/ir/native/PyDeserializerBuffer.hpp>
+#include <clp_ffi_py/PyObjectUtils.hpp>
+
+namespace clp_ffi_py::ir::native {
+auto DeserializerBufferReader::create(PyObject* input_stream, Py_ssize_t buf_capacity)
+        -> gsl::owner<DeserializerBufferReader*> {
+    PyObjectPtr<PyDeserializerBuffer> const py_deserializer_buffer{
+            PyDeserializerBuffer::create(input_stream, buf_capacity)
+    };
+    if (nullptr == py_deserializer_buffer) {
+        return nullptr;
+    }
+    return new DeserializerBufferReader{py_deserializer_buffer.get()};
+}
+
+auto DeserializerBufferReader::try_read(char* buf, size_t num_bytes_to_read, size_t& num_bytes_read)
+        -> clp::ErrorCode {
+    std::span<int8_t> dst_buf{clp::size_checked_pointer_cast<int8_t>(buf), num_bytes_to_read};
+    num_bytes_read = 0;
+    while (false == dst_buf.empty()) {
+        if (is_deserializer_buffer_empty() && false == fill_deserializer_buffer()) {
+            return clp::ErrorCode_EndOfFile;
+        }
+        auto const buffered_bytes_view{m_py_deserializer_buffer->get_unconsumed_bytes()};
+        auto const bytes_to_copy{
+                buffered_bytes_view.subspan(0, std::min(buffered_bytes_view.size(), dst_buf.size()))
+        };
+        std::copy(bytes_to_copy.begin(), bytes_to_copy.end(), dst_buf.begin());
+
+        // Commit read
+        auto const num_bytes_copied{bytes_to_copy.size()};
+        if (false
+            == m_py_deserializer_buffer->commit_read_buffer_consumption(
+                    static_cast<Py_ssize_t>(num_bytes_copied)
+            ))
+        {
+            throw ExceptionFFI(
+                    clp::ErrorCode_Failure,
+                    __FILE__,
+                    __LINE__,
+                    "`commit_read_buffer_consumption` failed"
+            );
+        }
+        num_bytes_read += num_bytes_copied;
+        dst_buf = dst_buf.last(dst_buf.size() - num_bytes_copied);
+    }
+
+    return clp::ErrorCode_Success;
+}
+
+auto DeserializerBufferReader::fill_deserializer_buffer() -> bool {
+    if (m_py_deserializer_buffer->try_read()) {
+        return true;
+    }
+    if (static_cast<bool>(
+                PyErr_ExceptionMatches(PyDeserializerBuffer::get_py_incomplete_stream_error())
+        ))
+    {
+        PyErr_Clear();
+        return false;
+    }
+    throw ExceptionFFI(
+            clp::ErrorCode_Failure,
+            __FILE__,
+            __LINE__,
+            "`DeserializerBufferReader::fill_deserializer_buffer` failed"
+    );
+}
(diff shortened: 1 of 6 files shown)


src/clp_ffi_py/ir/native/DeserializerBufferReader.hpp
```diff
@@ -0,0 +1,105 @@
...
+     * @return ErrorCode_EndOfFile if there is no more data to read.
+     * @throw ExceptionFFI that forwards `fill_deserializer_buffer`'s throw.
+     */
+    [[nodiscard]] auto try_read(char* buf, size_t num_bytes_to_read, size_t& num_bytes_read)
+            -> clp::ErrorCode override;
+
+    /**
+     * Tries to seek to the given position, relative to the beginning of the data.
+     * TODO: Implement this method when needed.
+     * @param pos
```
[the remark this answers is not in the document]
LinZhihao-723: I don't think we need to implement seek until we have a usecase

src/clp_ffi_py/ir/native/DeserializerBufferReader.hpp
```diff
@@ -0,0 +1,105 @@
...
+     * @return clp::ErrorCode_Unsupported
+     */
+    [[nodiscard]] auto try_seek_from_begin([[maybe_unused]] size_t pos) -> clp::ErrorCode override {
+        return clp::ErrorCode_Unsupported;
+    }
+
+    /**
+     * @param pos Returns the position of the read head in the buffer.
+     * @return ErrorCode_Success on success.
```
haiqi96: @kirkrodrigues if there is only one return value. I believe we can write
"@return ErrorCode_Success."?

src/clp_ffi_py/ir/native/DeserializerBufferReader.cpp
```diff
@@ -0,0 +1,83 @@
...
+namespace clp_ffi_py::ir::native {
+auto DeserializerBufferReader::create(PyObject* input_stream, Py_ssize_t buf_capacity)
+        -> gsl::owner<DeserializerBufferReader*> {
+    PyObjectPtr<PyDeserializerBuffer> const py_deserializer_buffer{
+            PyDeserializerBuffer::create(input_stream, buf_capacity)
+    };
+    if (nullptr == py_deserializer_buffer) {
+        return nullptr;
+    }
+    return new DeserializerBufferReader{py_deserializer_buffer.get()};
```
haiqi96, line 27: is there an implicit conversion from DeserializerBufferReader* to gsl::owner<DeserializerBufferReader*>?

Wonder why we don't make it explicit as https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/owning-memory.html

src/clp_ffi_py/ir/native/PyDeserializerBuffer.hpp
```diff
@@ -161,6 +161,19 @@ class PyDeserializerBuffer {
      */
     [[nodiscard]] auto test_streaming(uint32_t seed) -> PyObject*;
 
+    /**
+     * CPython-level factory function.
+     * @param input_stream
+     * @param buf_capacity
+     * @return a new reference of a `PyDeserializerBuffer` object that is initialized with the given
```
haiqi96, line 168: returns a pointer? unless reference and pointer are inter-changable terms in cpython

src/clp_ffi_py/ir/native/DeserializerBufferReader.hpp
```diff
@@ -0,0 +1,105 @@
...
+     * @param py_deserializer_buffer
+     */
+    explicit DeserializerBufferReader(PyDeserializerBuffer* py_deserializer_buffer)
+            : m_py_deserializer_buffer{py_deserializer_buffer} {
+        Py_INCREF(py_deserializer_buffer);
+    }
+
+    // Methods
+    /**
+     * Fills the underlying deserializer buffer by calling its `try_read`.
```
haiqi96, line 88: nit: wonder if we need to discuss about the implementation details. but don't think it hurts for me.

src/clp_ffi_py/ir/native/PyDeserializerBuffer.cpp
```diff
@@ -376,6 +376,21 @@ auto PyDeserializerBuffer::test_streaming(uint32_t seed) -> PyObject* {
 PyObjectStaticPtr<PyTypeObject> PyDeserializerBuffer::m_py_type{nullptr};
 PyObjectStaticPtr<PyObject> PyDeserializerBuffer::m_py_incomplete_stream_error{nullptr};
 
+auto PyDeserializerBuffer::create(PyObject* input_stream, Py_ssize_t buf_capacity)
+        -> PyDeserializerBuffer* {
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-cstyle-cast)
```
haiqi96, line 381: do we still need this since next line doesn't have any cast?

src/clp_ffi_py/ir/native/PyDeserializerBuffer.hpp
LinZhihao-723: `new reference` is a CPython terminology, meaning that it creates a new Python object through C/C++ level API

LinZhihao-723: https://pythonextensionpatterns.readthedocs.io/en/latest/refcount.html#new-references

src/clp_ffi_py/ir/native/DeserializerBufferReader.cpp
LinZhihao-723: `gsl::owner<T>` is `T` itself, so there's no implicit conversion. It is only used to silent clang-tidy warnings, as only owner typed pointer can be the parameter of `new` or `delete` operators.
As showing in this site, clang-tidy doesn't always use `static_cast` to cast a pointer type to the owner type. And also the static cast won't be able to do any check as the owner type doesn't actually enforce anything. It's just a hint for the linter
So I would just keep it directly returning without any type casting

src/clp_ffi_py/ir/native/DeserializerBufferReader.hpp
LinZhihao-723: Hmmm, since it's a private method, I think it's fine...
But it might be a good practice for doc string...
@kirkrodrigues Any comments?

src/clp_ffi_py/ir/native/PyDeserializerBuffer.cpp
LinZhihao-723: `PyObject_New` is a macro and it uses C-style casting (welcome to CPython, lol)

src/clp_ffi_py/ir/native/DeserializerBufferReader.hpp
kirkrodrigues: I would write:

`@return ErrorCode_Success always.`

src/clp_ffi_py/ir/native/PyDeserializerBuffer.hpp
haiqi96: sounds good.

src/clp_ffi_py/ir/native/DeserializerBufferReader.hpp
kirkrodrigues: Personally, I try to keep all docstrings (even private method ones) agnostic to their implementation; but as you said, for private methods, that isn't a requirement.

src/clp_ffi_py/ir/native/PyDeserializerBuffer.cpp
haiqi96: GG, learned

haiqi96 (approved):
Looks good to me

PR #90, opened 2024-11-16
perf: Optimize `KeyValuePairLogEvent.to_dict` to directly convert `clp::KeyValuePairLogEvent` to `PyDictObject` via CPython APIs.

LinZhihao-723:
# Description

In #81 we introduced `PyKeyValuePairLogEvent` as a wrapper of `clp::ffi::KeyValuePairLogEvent`. However, we didn't properly implement `to_dict` method in that PR, which works as `clp::ffi::KeyValuePairLogEvent` -> `nlohmann::json` -> JSON string -> Python dictionary (by parsing the JSON string). This is very inefficient as explained in #81.
This PR implements  `clp::ffi::KeyValuePairLogEvent` -> Python dictionary conversion properly by using the sub-schema tree DFS. By benchmarking against elastic search's log dataset [here](https://zenodo.org/records/10516227), it is around 3x faster than the workflow above.

# Validation performed

- Ensure the current unit tests all passed, which cover `clp::ffi::KeyValuePairLogEvent` -> Python dictionary conversion.
- Tested within the beta branch #86
- Ran with address sanitizer to ensure no warnings/errors generated
- Ensure workflows passed

## Summary by CodeRabbit

- **New Features**
	- Enhanced serialization and deserialization of log events, allowing for better conversion of log data into Python dictionary objects.
	- Introduced a new method, `to_dict()`, in the key-value pair log event class to facilitate the conversion of log events to Python dictionaries.

- **Bug Fixes**
	- Improved error handling during serialization, providing more informative error messages for various failure conditions.


src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -314,6 +488,157 @@
...
+    auto const key_name{node.get_key_name()};
+    if (false == optional_val.has_value()) {
+        PyObjectPtr<PyObject> const empty_dict{PyDict_New()};
+        return 0
+               == PyDict_SetItemString(
+                       py_reinterpret_cast<PyObject>(dict),
+                       key_name.data(),
+                       empty_dict.get()
+               );
+    }
```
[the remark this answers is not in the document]
LinZhihao-723: No need. `PyObjectPtr` handles that already

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -325,6 +650,30 @@
...
+        return serialize_node_id_value_pair_to_py_dict(
+                schema_tree,
+                schema_subtree_bitmap_result.value(),
+                node_id_value_pairs
+        );
+    } catch (clp::TraceableException& ex) {
+        handle_traceable_exception(ex);
+        return nullptr;
+    }
+}
```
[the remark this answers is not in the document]
LinZhihao-723: There should be no other std::exception to catch

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -314,6 +488,157 @@ auto convert_py_dict_to_key_value_pair_log_event(PyDictObject* py_dict
...
+    auto const result{
+            val.is<FourByteEncodedTextAst>()
+                    ? val.get_immutable_view<FourByteEncodedTextAst>().decode_and_unparse()
+                    : val.get_immutable_view<EightByteEncodedTextAst>().decode_and_unparse()
+    };
+    if (false == result.has_value()) {
+        PyErr_SetString(PyExc_RuntimeError, "Failed to deserialize CLP encoded text AST");
+    }
+    return result;
+}
```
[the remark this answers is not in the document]
LinZhihao-723: This is ensured by the caller

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -65,6 +78,149 @@ class IrUnitHandler {
...
+            PyDictObject* parent
+    ) -> std::optional<PyDictSerializationIterator> {
+        if (schema_tree_node->is_root() && nullptr != parent) {
+            PyErr_SetString(
+                    PyExc_RuntimeError,
+                    "KeyValuePairLogEvent.to_dict(): Root node cannot have a parent"
+            );
+            return std::nullopt;
+        }
+        if (false == schema_tree_node->is_root() && nullptr == parent) {
```
haiqi96, line 108: Nit:

Those two if clause and be merged into one as. but then we need a clever error message.. so it's really just a "FYI"

if ((schema_tree_node->is_root()) != (nullptr == parent))

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -65,6 +78,149 @@ class IrUnitHandler {
...
+     */
+    [[nodiscard]] auto has_next_child_schema_tree_node() const -> bool {
+        return m_child_schema_tree_node_it != m_child_schema_tree_nodes.end();
+    }
+
+    /**
+     * Gets the next child schema tree node and advances the iterator.
+     * @return The next child schema tree node.
+     */
+    [[nodiscard]] auto get_next_child_schema_tree_node() -> SchemaTree::Node::id_t {
```
haiqi96: how about
```suggestion
    [[nodiscard]] auto get_next_child_schema_tree_node_id() -> SchemaTree::Node::id_t {
```

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -65,6 +78,149 @@ class IrUnitHandler {
...
+            : m_schema_tree_node{schema_tree_node},
+              m_child_schema_tree_nodes{std::move(child_schema_tree_nodes)},
+              m_child_schema_tree_node_it{m_child_schema_tree_nodes.cbegin()},
+              m_parent{parent},
+              m_py_dict{std::move(py_dict)} {}
+
+    SchemaTree::Node const* m_schema_tree_node;
+    std::vector<SchemaTree::Node::id_t> m_child_schema_tree_nodes;
+    std::vector<SchemaTree::Node::id_t>::const_iterator m_child_schema_tree_node_it;
+    PyDictObject* m_parent;
```
haiqi96: how about m_parent_py_dict?

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -65,6 +78,149 @@ class IrUnitHandler {
...
+    [[nodiscard]] auto get_next_child_schema_tree_node() -> SchemaTree::Node::id_t {
+        return *(m_child_schema_tree_node_it++);
+    }
+
+    /**
+     * Adds the underlying Python dictionary into the parent.
+     * @return true on success.
+     * @return false on failure with the relevant Python exception and error set.
+     */
+    [[nodiscard]] auto add_to_parent() -> bool {
```
haiqi96: how about something like add_to_parent_py_dict

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -314,6 +488,157 @@ auto convert_py_dict_to_key_value_pair_log_event(PyDictObject* py_dict
...
+    std::stack<DfsIterator> dfs_stack;
+    auto optional_root_iterator
+            = DfsIterator::create(&schema_tree.get_root(), schema_subtree_bitmap, nullptr);
+    if (false == optional_root_iterator.has_value()) {
+        return nullptr;
+    }
+    dfs_stack.emplace(std::move(optional_root_iterator.value()));
+
+    while (false == dfs_stack.empty()) {
+        auto& top{dfs_stack.top()};
```
haiqi96: nit: instead of top, how about something like "next_node" or "top_node". Would make it easier to read for people who didn't write this code

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -314,6 +488,157 @@ auto convert_py_dict_to_key_value_pair_log_event(PyDictObject* py_dict
...
+            break;
+        case SchemaTree::Node::Type::UnstructuredArray: {
+            auto const decoded_result{decode_as_encoded_text_ast(val)};
+            if (false == decoded_result.has_value()) {
+                return false;
+            }
+            py_value.reset(py_utils_parse_json_str(decoded_result.value()));
+            break;
+        }
+        case SchemaTree::Node::Type::Obj:
```
haiqi96, line 607: Correct me if wrong, but I thought in this PR's logic, a node with Type::Obj  should not end up enterring insert_kv_pair_into_py_dict?

haiqi96 (commented):
In general, the code makes sense to me.

I have one comment on PyDictSerializationIterator though. Currently it's acting as two roles;
1. handle the "next_node" for dfs path
2. Maintain and construct the dictionary in a bottom up manner (if my understanding is correct)

The 1st use case makes total sense. However, the 2nd use case is not very intuitive just by looking at the class name, and it took me sometime to digest the code. 

If I were to write this code, I might consider factor out the dictionary maintaining code from the iterator. I am not suggesting a major refactor of this code though. Can we have some high level comments for the class to describe what it does?

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -65,6 +78,149 @@ class IrUnitHandler {
...
+                       py_reinterpret_cast<PyObject>(m_py_dict.get())
+               );
+    }
+
+    /**
+     * Releases the underlying Python dictionary as the root dictionary to return.
+     * @return The released Python dictionary on success.
+     * @return nullptr on failure with the relevant Python exception and error set.
+     */
+    [[nodiscard]] auto release_root() -> PyDictObject* {
```
haiqi96, line 188: Thinking if this can be simplified. Pending on the offline discussion

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
LinZhihao-723: if the type is an object and it's empty, like `{}`, it can be a leaf node in the flatten tree. For example:
```
{"key": {"inner_key": {}}}
```

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
LinZhihao-723: yeah basically it's an XOR here. I split them out cuz the error messages should be different...

LinZhihao-723:
> In general, the code makes sense to me.
> 
> I have one comment on PyDictSerializationIterator though. Currently it's acting as two roles;
> 
> 1. handle the "next_node" for dfs path
> 2. Maintain and construct the dictionary in a bottom up manner (if my understanding is correct)
> 
> The 1st use case makes total sense. However, the 2nd use case is not very intuitive just by looking at the class name, and it took me sometime to digest the code.
> 
> If I were to write this code, I might consider factor out the dictionary maintaining code from the iterator. I am not suggesting a major refactor of this code though. Can we have some high level comments for the class to describe what it does?

As discussed offline, the reason to let the iterator to maintain the dictionary is to simplify the scenario when existing on errors. If an error happens, the stack will be destructed, and thus, recursively destruct all iterators inside the stack, including the dictionaries (either ones still in the stack or those already added to parent dicts). Therefore, the conclusion is we will keep it this way for the coming release.

LinZhihao-723:
The naming suggestions make sense to me. I will fix them in the coming commits @haiqi96

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
haiqi96: I see, thanks

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
haiqi96: Got it, makes sense

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
LinZhihao-723: strictly speaking it's just an iterator but not a node...

LinZhihao-723: How about `curr_stack_top`?
Any other suggestions? @kirkrodrigues

kirkrodrigues: How about `dfs_stack_top`?

haiqi96: >strictly speaking it's just an iterator but not a node...

I guess it depends on how someone visualize the algorithm... I fell it's ok to use dfs_stack_top though, anyway I managed to understand what's going on before reading the IRv2 spec, so the name shouldn't be a huge issue.

LinZhihao-723: Sure, gonna use `dfs_stack_top`

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -65,6 +78,149 @@ class IrUnitHandler {
...
+     */
+    [[nodiscard]] auto has_next_child_schema_tree_node_id() const -> bool {
+        return m_child_schema_tree_node_it != m_child_schema_tree_nodes.end();
+    }
+
+    /**
+     * Gets the next child schema tree node and advances the iterator.
+     * @return The next child schema tree node.
+     */
+    [[nodiscard]] auto get_next_child_schema_tree_node() -> SchemaTree::Node::id_t {
```
haiqi96: You might have added _id to the wrong function, lol

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -65,6 +78,149 @@ class IrUnitHandler {
...
+    /**
+     * @return Whether there are more child schema tree nodes to traverse.
+     */
+    [[nodiscard]] auto has_next_child_schema_tree_node() const -> bool {
+        return m_child_schema_tree_node_it != m_child_schema_tree_nodes.end();
+    }
+
+    /**
+     * Gets the next child schema tree node and advances the iterator.
+     * @return The next child schema tree node.
```
haiqi96: One last nit
```suggestion
     * Gets the id of the next child schema tree node and advances the iterator.
     * @return The id of the next child schema tree node.
```

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
LinZhihao-723: yeah, mb. Fixed now

haiqi96 (approved):
Looks good to me.

Maybe we can remove "directly" from the title?

feat: Add support for converting clp::ffi::KeyValuePairLogEvent to Python dictionary.

LinZhihao-723:
> Looks good to me.
> 
> Maybe we can remove "directly" from the title?
> 
> feat: Add support for converting clp::ffi::KeyValuePairLogEvent to Python dictionary.

How about:
```
perf: Optimize `KeyValuePairLogEvent.to_dict` method to perform direct `clp::KeyValuePairLogEvent`-to-`PyDictObject` conversion via CPython APIs.
```
- Using `perf` since this is actually a performance improvement but not introducing new features
- Using "direct" to emphasize the conversion doesn't have any intermediate stage
- Adding "via CPython APIs" to indicate the conversion happens purely in CPython (ffi layer)

Do you think this makes sense? @kirkrodrigues

kirkrodrigues:
Sure. How about a slightly shorter one:

```
perf: Optimize `KeyValuePairLogEvent.to_dict` to directly convert `clp::KeyValuePairLogEvent` to `PyDictObject` via CPython APIs.
```

LinZhihao-723:
> Sure. How about a slightly shorter one:
> 
> ```
> perf: Optimize `KeyValuePairLogEvent.to_dict` to directly convert `clp::KeyValuePairLogEvent` to `PyDictObject` via CPython APIs.
> ```

thx!

PR #94, opened 2024-11-23
feat: Add Python class `Deserializer` to deserialize log events from CLP key-value pair IR stream.

LinZhihao-723:
# Description

This PR adds a new class `Deserializer` to support deserializing log events from CLP key-value pair IR stream. The `Deserializer` is designed to read from a Python IO[byte] object as the input stream, and emit `KeyValueLogEvent` instances as the deserialized output.

# Validation performed

- Ensure all unit tests passed, with the newly added serder tests.
- Ensure workflow passed.
- Ran with address sanitizer to ensure no memory errors
- Tested in the beta branch #86 

## Release Notes

- **New Features**
	- Introduced the `PyDeserializer` class for enhanced deserialization of CLP key-value pair IR streams.
	- Added a `Deserializer` class to the public API, allowing users to easily access deserialization functionalities.
	- Implemented a new method in `PyKeyValuePairLogEvent` for streamlined object creation from log events.

- **Tests**
	- Added a comprehensive testing framework for serialization and deserialization processes, ensuring robust validation of functionality.


src/clp_ffi_py/ir/native/PyDeserializer.hpp
```diff
@@ -0,0 +1,255 @@
...
+    [[nodiscard]] auto is_stream_complete() const -> bool { return m_end_of_stream_reached; }
+
+    // Variables
+    PyObject_HEAD;
+    bool m_end_of_stream_reached;
+    bool m_allow_incomplete_stream;
+    // NOLINTBEGIN(cppcoreguidelines-owning-memory)
+    gsl::owner<DeserializerBufferReader*> m_deserializer_buffer_reader;
+    gsl::owner<Deserializer*> m_deserializer;
+    gsl::owner<clp::ffi::KeyValuePairLogEvent*> m_deserialized_log_event;
```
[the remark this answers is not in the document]
LinZhihao-723: This is intentional for CPython

src/clp_ffi_py/ir/native/PyDeserializer.hpp
```diff
@@ -0,0 +1,255 @@
...
+    PyDeserializer() = delete;
+
+    // Delete copy/move constructors and assignments
+    PyDeserializer(PyDeserializer const&) = delete;
+    PyDeserializer(PyDeserializer&&) = delete;
+    auto operator=(PyDeserializer const&) -> PyDeserializer& = delete;
+    auto operator=(PyDeserializer&&) -> PyDeserializer& = delete;
+
+    // Destructor
+    ~PyDeserializer() = default;
```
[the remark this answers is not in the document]
LinZhihao-723: It's for CPython.

src/clp_ffi_py/ir/native/PyDeserializer.cpp
```diff
@@ -0,0 +1,281 @@
...
+            PyErr_Format(
+                    PyExc_RuntimeError,
+                    cDeserializerCreateErrorFormatStr.data(),
+                    deserializer_result.error().message().c_str()
+            );
+            return false;
+        }
+        m_deserializer = new clp::ffi::ir_stream::Deserializer<PyDeserializer::IrUnitHandler>{
+                std::move(deserializer_result.value())
+        };
```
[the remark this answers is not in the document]
LinZhihao-723: Same, this is for CPython

src/clp_ffi_py/ir/native/PyDeserializer.cpp
```diff
@@ -0,0 +1,281 @@
...
+    return add_python_type(get_py_type(), "Deserializer", py_module);
+}
+
+auto PyDeserializer::handle_log_event(clp::ffi::KeyValuePairLogEvent&& log_event) -> IRErrorCode {
+    if (has_unreleased_deserialized_log_event()) {
+        release_deserialized_log_event();
+    }
+    m_deserialized_log_event = new clp::ffi::KeyValuePairLogEvent{std::move(log_event)};
+    return IRErrorCode::IRErrorCode_Success;
+}
```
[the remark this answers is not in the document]
LinZhihao-723: Same

src/clp_ffi_py/ir/native/PyDeserializer.cpp
```diff
@@ -0,0 +1,281 @@
...
+                );
+            }
+        }
+    } catch (clp::TraceableException& exception) {
+        handle_traceable_exception(exception);
+        return nullptr;
+    }
+
+    Py_RETURN_NONE;
+}
```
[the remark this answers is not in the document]
LinZhihao-723: There should be no infinite loop.

tests/test_ir/test_serder.py
```diff
@@ -0,0 +1,175 @@
+from pathlib import Path
+from typing import Any, Dict, IO, List, Optional
+
+from smart_open import open  # type: ignore
```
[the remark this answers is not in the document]
LinZhihao-723: `smart_open.open` here should be designed to replace build-in `open`

clp_ffi_py/ir/__init__.py
```diff
@@ -8,6 +8,7 @@
 __all__: List[str] = [
     "Decoder",  # native_deprecated
     "DecoderBuffer",  # native_deprecated
+    "Deserializer",  # Native
```
haiqi96: why captial 'N' while others have small 'n'

src/clp_ffi_py/ir/native/PyDeserializer.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+namespace {
+/**
+ * Callback of `PyDeserializer`'s `__init__` method:
+ */
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays)
+PyDoc_STRVAR(
+        cPyDeserializerDoc,
+        "Deserializer for deserializing CLP key-value pair IR streams.\n"
+        "This class deserializes log events from a CLP key-value pair IR stream.\n\n"
```
haiqi96: Nit：
```suggestion
        "This class deserializes a CLP key-value pair IR stream into log events.\n\n"
```

src/clp_ffi_py/ir/native/PyDeserializer.cpp
```diff
@@ -0,0 +1,281 @@
...
+        "__init__(self, input_stream, buffer_capacity=65536, allow_incomplete_stream=False)\n\n"
+        "Initializes a :class:`Deserializer` instance with the given inputs. Note that each"
+        " object should only be initialized once. Double initialization will result in a memory"
+        " leak.\n\n"
+        ":param input_stream: Input stream that contains serialized CLP IR.\n"
```
haiqi96: Perhaps you don't need to double mentioning "input stream" since the variable name already explains itself?
```suggestion
        ":param input_stream: serialized CLP IR stream.\n"
```

src/clp_ffi_py/ir/native/PyDeserializer.cpp
```diff
@@ -0,0 +1,281 @@
...
+        ":type input_stream: IO[bytes]\n"
+        ":param buffer_capacity: The capacity used to initialize the underlying read buffer.\n"
+        ":type buffer_capacity: int\n"
+        ":param allow_incomplete_stream: If set to `True`, an incomplete CLP IR stream is not"
```
haiqi96, line 45: Nit: I am not sure if "truncated stream" would be better if incomplete refers to "formatly correct but cut off unexpectedly at the end"
Technically, incomplete can also mean that you somehow lose a tag at the middle of the stream.

@kirkrodrigues any comments?

src/clp_ffi_py/ir/native/PyDeserializer.cpp
```diff
@@ -0,0 +1,281 @@
...
+        cPyDeserializerDoc,
+        "Deserializer for deserializing CLP key-value pair IR streams.\n"
+        "This class deserializes log events from a CLP key-value pair IR stream.\n\n"
+        "__init__(self, input_stream, buffer_capacity=65536, allow_incomplete_stream=False)\n\n"
+        "Initializes a :class:`Deserializer` instance with the given inputs. Note that each"
+        " object should only be initialized once. Double initialization will result in a memory"
+        " leak.\n\n"
+        ":param input_stream: Input stream that contains serialized CLP IR.\n"
+        ":type input_stream: IO[bytes]\n"
+        ":param buffer_capacity: The capacity used to initialize the underlying read buffer.\n"
```
haiqi96: Will the reader buffer size change?
if not, I guess "initialize" is not important. we can just write
```suggestion
        ":param buffer_capacity: The capacity of the underlying read buffer.\n"
```

src/clp_ffi_py/ir/native/PyDeserializer.cpp
```diff
@@ -0,0 +1,281 @@
...
+PyDeserializer_init(PyDeserializer* self, PyObject* args, PyObject* keywords) -> int;
+
+/**
+ * Callback of `PyDeserializer`'s `deserialize_log_event`.
+ */
+PyDoc_STRVAR(
+        cPyDeserializerDeserializeLogEventDoc,
+        "deserialize_log_event(self)\n"
+        "--\n\n"
+        "Deserializes the IR stream until the next log event has been deserialized.\n\n"
```
haiqi96: Tbh, the sentence looks confusing to me.
How about?
```suggestion
        "Deserializes the next log event from the IR stream.\n\n"
```

src/clp_ffi_py/ir/native/PyDeserializer.cpp
```diff
@@ -0,0 +1,281 @@
...
+        ":return:\n"
+        "     - The next deserialized log event from the IR stream.\n"
+        "     - None if there are no more log events in the stream.\n"
```
haiqi96, line 62: do we want to mention the behavior for incomplete stream?

src/clp_ffi_py/ir/native/PyDeserializer.cpp
```diff
@@ -0,0 +1,281 @@
...
+                        PyKeyValuePairLogEvent::create(release_deserialized_log_event())
+                );
+            }
+        }
+    } catch (clp::TraceableException& exception) {
+        handle_traceable_exception(exception);
+        return nullptr;
+    }
+
+    Py_RETURN_NONE;
```
haiqi96, line 259: only `break` will reach this line, what's your thought on directly returning Py_RETURN_NONE instead of calling `break`?

src/clp_ffi_py/ir/native/PyDeserializer.cpp
```diff
@@ -0,0 +1,281 @@
...
+        handle_traceable_exception(exception);
+        return false;
+    }
+
+    return true;
+}
+
+auto PyDeserializer::deserialize_log_event() -> PyObject* {
+    try {
+        while (false == is_stream_complete()) {
```
haiqi96: Nit:
I feel using `is_stream_complete()` is more confusing then directly using m_end_of_stream_reached.

Reading the code, what i feel is that you want to keep going if the stream is "incomplete", but the intension is to keep going as long as the stream hasn't ended.

If you agree with me, maybe we don't need `is_stream_complete` anymore.

Edit: or maybe you should name it as `is_stream_completed` with the extra "d"

haiqi96 (commented):
First batch of comments

src/clp_ffi_py/ir/native/PyDeserializer.cpp
kirkrodrigues: "truncated" sounds reasonable to me.

src/clp_ffi_py/ir/native/PyDeserializer.cpp
haiqi96: my bad, ignore this comment

src/clp_ffi_py/ir/native/PyDeserializer.cpp
```diff
@@ -0,0 +1,281 @@
...
+                    m_deserializer->deserialize_next_ir_unit(*m_deserializer_buffer_reader)
+            };
+            if (ir_unit_type_result.has_error()) {
+                if (false == handle_incomplete_ir_error(ir_unit_type_result.error())) {
+                    return nullptr;
+                }
+                break;
+            }
+            auto const ir_unit_type{ir_unit_type_result.value()};
+            if (IrUnitType::LogEvent == ir_unit_type && has_unreleased_deserialized_log_event()) {
```
haiqi96: Just to help me understand the code,

what would be a case where
IrUnitType::LogEvent == ir_unit_type is True, but has_unreleased_deserialized_log_event is False

src/clp_ffi_py/ir/native/PyDeserializer.cpp
```diff
@@ -0,0 +1,281 @@
...
+    auto* type{py_reinterpret_cast<PyTypeObject>(PyType_FromSpec(&PyDeserializer_type_spec))};
+    m_py_type.reset(type);
+    if (nullptr == type) {
+        return false;
+    }
+    return add_python_type(get_py_type(), "Deserializer", py_module);
+}
+
+auto PyDeserializer::handle_log_event(clp::ffi::KeyValuePairLogEvent&& log_event) -> IRErrorCode {
+    if (has_unreleased_deserialized_log_event()) {
```
haiqi96, line 273: In normal operation, if no code runs into error, do you see a possibility that when this method is called, there is unrealised deserialized_log_event?

src/clp_ffi_py/ir/native/PyDeserializer.hpp
```diff
@@ -0,0 +1,255 @@
...
+    auto clean() -> void {
+        delete m_deserializer;
+        delete m_deserializer_buffer_reader;
+        if (has_unreleased_deserialized_log_event()) {
+            release_deserialized_log_event();
+        }
+    }
+
+    /**
+     * Deserializes to the next key value pair log event from the IR stream.
```
haiqi96: How about
```suggestion
     * Deserializes the next key value pair log event from the IR stream.
```

src/clp_ffi_py/ir/native/PyDeserializer.hpp
```diff
@@ -0,0 +1,255 @@
...
+        return released;
+    }
+
+    /**
+     * Handles the incomplete IR error returned from `Deserializer::deserialize_next_ir_unit`.
+     * @param err
+     * @return true if the error is resolved on success.
+     * @return false with the relevant Python exception and error set on failures:
+     * - The error doesn't indicate an incomplete IR stream.
+     * - Incomplete IR stream is not allowed.
```
haiqi96: What do these two sentences mean?
it's not very clear to me, can you elaborate?

tests/test_ir/test_serder.py
```diff
@@ -0,0 +1,175 @@
...
+        jsonl_path: Path,
+    ) -> Path:
+        ir_stream_path: Path = LOG_DIR / Path(
+            f"{self.ir_stream_path_prefix}.{str(jsonl_path.stem)}.{self.ir_stream_path_postfix}"
+        )
+        if ir_stream_path.exists():
+            ir_stream_path.unlink()
+        return ir_stream_path
+
+    def test_serder(self) -> None:
```
haiqi96, line 106: Where is this test method called?

src/clp_ffi_py/ir/native/PyDeserializer.cpp
LinZhihao-723: `incomplete_stream` is the term we used since IR v1. The exception we through for "truncated" stream is also called `incomplete stream error`. This name is determined in this PR: #18. Prefer to keep it unchanged for the coming release, and have further discussions whether to improve it or not in future releases.

haiqi96: I see, if it's something that's determined previously then let's keep it for now.

clp_ffi_py/ir/__init__.py
LinZhihao-723: that's a mistake. Will fix it in the coming commits

src/clp_ffi_py/ir/native/PyDeserializer.cpp
LinZhihao-723: It is written this way since the stream may also contain other IR units. This method deserializes all IR units until the next log event IR unit has been deserialized

src/clp_ffi_py/ir/native/PyDeserializer.cpp
LinZhihao-723: I think we don't have to explicitly speak it out:
- The default config is `allow_incomplete_stream` set to false, meaning that an incomplete stream should be treated as an exception
- If it's set to true, "there are no more log events in the stream" will correctly define this situation, and returning None without exceptions is naturally expected

haiqi96: Sure.

src/clp_ffi_py/ir/native/PyDeserializer.cpp
haiqi96: Do customers need to be aware of the IR units? If so then I agree with you.

LinZhihao-723: At this point no. Sure, let's take ur suggestion for this release

src/clp_ffi_py/ir/native/PyDeserializer.cpp
LinZhihao-723: thinking again. I think we don't need to check the deserialized IR unit type at all. We just need to check whether it returns an error or not.
But there could be cases where the return IR unit type is `LogEvent` without unreleased log event buffered. If we implement native query, only matched log event will be stored.

src/clp_ffi_py/ir/native/PyDeserializer.cpp
LinZhihao-723: No. It's only possible if an error happened

src/clp_ffi_py/ir/native/PyDeserializer.hpp
LinZhihao-723: This function is added to "swallow" incomplete stream error if needed.

- `The error doesn't indicate an incomplete IR stream.`: Only one particular `std::error_code` indicates an incomplete stream error. If the given error isn't an incomplete stream error, we should return false as we can't handle this error.
- `Incomplete IR stream is not allowed.`: If the error code does indicate an incomplete stream error, but it is not allowed in the current deserializer, we should also return false as this is not the error we can swallow.

tests/test_ir/test_serder.py
LinZhihao-723: Every derived class will call this method implicitly.
Python unit test framework will call every member methods that start with `test_`

src/clp_ffi_py/ir/native/PyDeserializer.cpp
LinZhihao-723: Sure, using `is_stream_completed` makes sense.
Just realized we used `is_stream_completed` in core as well: https://github.com/y-scope/clp/blob/main/components/core/src/clp/ffi/ir_stream/Deserializer.hpp#L105

src/clp_ffi_py/ir/native/PyDeserializer.cpp
haiqi96: In that case, shouldn't we flag the error instead of silently deleting the old log event?

tests/test_ir/test_serder.py
haiqi96: I see, fair enough

src/clp_ffi_py/ir/native/PyDeserializer.cpp
LinZhihao-723: The error must already be returned and handled before control flow reaches here.

Also this is more like a sanity check. This is not like unique pointer where overwriting the pointer will automatically free the underlying object. Without releasing non-empty log event it would lead to memory leak.

LinZhihao-723: Added a comment to explain why this is required

src/clp_ffi_py/ir/native/PyDeserializer.cpp
```diff
@@ -0,0 +1,286 @@
...
+                break;
+            }
+            if (IrUnitType::LogEvent != ir_unit_type_result.value()
+                || false == has_unreleased_deserialized_log_event())
+            {
+                continue;
+            }
+            return py_reinterpret_cast<PyObject>(
+                    PyKeyValuePairLogEvent::create(release_deserialized_log_event())
+            );
```
LinZhihao-723: Actually it might be better if the return IR unit type is checked. Re-write it this way, do u think it's better? @haiqi96

src/clp_ffi_py/ir/native/PyDeserializer.cpp
```diff
@@ -0,0 +1,286 @@
...
+            nullptr
+    };
+
+    // If the argument parsing fails, `self` will be deallocated. We must reset all pointers to
+    // nullptr in advance, otherwise the deallocator might trigger segmentation fault.
+    self->default_init();
+
+    PyObject* input_stream{};
+    Py_ssize_t buffer_capacity{PyDeserializer::cDefaultBufferCapacity};
+    int allow_incomplete_stream{0};
```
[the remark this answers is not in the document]
LinZhihao-723: Good catch. This is actually an issue in `PyDeserializerBuffer`. I will send another PR to fix it.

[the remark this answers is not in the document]
LinZhihao-723: #97

src/clp_ffi_py/ir/native/PyDeserializer.cpp
haiqi96: Thanks for adding the comment.

My concern was that, while people reading the code, they might assume that such sanity check represents a possible run time situation in the code, which will make it harder to reason through the code.
A comment clarifying the purpose of such check would be very helpful

src/clp_ffi_py/ir/native/PyDeserializer.cpp
haiqi96: >But there could be cases where the return IR unit type is LogEvent without unreleased log event buffered. If we implement native query, only matched log event will be stored.

Do we expect the native query to be implemented soon? If it's not in the near future, I think we don't have to prepare for this case in the code, as it would cause some confusions.

It makes more sense to me to add the check for has_unreleased_deserialized_log_event() only when it is needed, or at least we should add a comment.

src/clp_ffi_py/ir/native/PyDeserializer.hpp
```diff
@@ -0,0 +1,257 @@
...
```
haiqi96: Now I get what you mean, but I feel there might be better way to phrase it. Here is what I can think of, @kirkrodrigues can you give some input?
```suggestion
     * @return false with the relevant Python exception and error set on the following cases:
     * - The error doesn't indicate an incomplete IR stream.
     * - The error indicates an incomplete IR stream, and `allow_incomplete_stream` == `false`.
```

src/clp_ffi_py/ir/native/PyDeserializer.cpp
LinZhihao-723: I have refactored this piece of code.
In general, I think:
- We should check the returned IR unit type as it's the expected usage of the core deserializer API
- We should also ensure there is an unreleased deserialized log event before we try to "release" this pointer

That said, I think both checks are necessary, but we can write it so that it doesn't look confusing.

LinZhihao-723: Disuccsed offline. The latest commit should reflect our conclusion:
- We should keep both check as discussed above
- We should check `has_unreleased_deserialized_log_event` to meet the requirement that the ownership must be validated by the caller before calling `release_deserialized_log_event`
- If the validation failed, we should throw a Python exception instead of `continue`. `continue` logic should be implemented in the PR which native queries are implemented

src/clp_ffi_py/ir/native/PyDeserializer.hpp
kirkrodrigues: I have some suggestions but the function's name is a bit confusing. If the error from the deserializer *wasn't* an incomplete IR error, then why is this method trying to handle it? Can we pick a name like `handle_deserializer_error` or `handle_deserialization_error`?

LinZhihao-723: This function is supposed to only handle incomplete IR errors. However, we have to check the error code to see if it's an incomplete error. So do u think it makes more sense to either:
- Rename this function to `handle_deserializer_error` and call incomplete IR error handler inside
- Check the error code before calling this function, and only call it when the error code indicates incomplete IR error

Notice that at this moment we don't handle any other errors except incomplete IR error. For any other errors we just need to return by setting the Python exception

kirkrodrigues: > Rename this function to handle_deserializer_error and call incomplete IR error handler inside

If you're renaming `handle_incomplete_ir_error` to `handle_deserializer_error`, then what do you mean by "call incomplete IR error handler inside"?

> Notice that at this moment we don't handle any other errors except incomplete IR error. For any other errors we just need to return by setting the Python exception

I would argue that "setting the Python exception" *is* handling the error. When we say we've handled all errors, it doesn't mean that every error is resolvable. It just means that this layer of the software does something for every error instead of crashing or invoking undefined behaviour. The docstring is what clarifies the behaviour for each error.

LinZhihao-723: hmmm, from the code behaviour I think we should check the error code before calling this function. So far we are only trying to resolve incomplete IR errors. For the rest of errors we should just throw them to users.
But I don't agree on the point that "setting the Python exception" is "handling errors." It's just like u're programming in only C++ and u throw whatever exception occurred, and we're arguing "throwing" == "handling" (the only difference is our exception cross ffi layer). Or maybe we can be more clear rather than using the word "handle". Based on the discussion, do u have better suggestion?

kirkrodrigues: If "handle" doesn't imply doing something with the error, then what is the difference between "resolve" and "handle"? Let's say we wanted to set a Python exception for every error, what would you call that method?

LinZhihao-723: - This function is not just set a Python exception for the given error. For incomplete IR error, we might just swallow it if users don't care
- If the "handler" is just to set an error, I would probably call it `set_python_exception(error_code)`, or as chatgpt suggested `map_error_to_exception`

kirkrodrigues: > This function is not just set a Python exception for the given error. For incomplete IR error, we might just swallow it if users don't care

Doesn't really answer my question of what is the difference between "handle" and "resolve".

Anyway, it's probably best you check the error outside the method.

src/clp_ffi_py/ir/native/PyDeserializer.cpp
```diff
@@ -0,0 +1,297 @@
...
+                            PyExc_RuntimeError,
+                            cDeserializerDeserializeNextIrUnitErrorFormatStr.data(),
+                            err.message().c_str()
+                    );
+                    return nullptr;
+                }
+                if (false == handle_incomplete_stream_error()) {
+                    return nullptr;
+                }
+                break;
```
haiqi96, line 233: what if we simply do this and remove the handle_incomplete_stream_error() method? The error handling should be the same but there will be a difference on how we report the error, not sure if it is an issue.
```suggestion
                if (std::errc::result_out_of_range == err && m_allow_incomplete_stream) {
                    handle_end_of_stream();
                    break;
                }
                PyErr_Format(
                    PyExc_RuntimeError,
                    cDeserializerDeserializeNextIrUnitErrorFormatStr.data(),
                    err.message().c_str()
                );
                return nullptr;
```
```

LinZhihao-723: I'm afraid the error reporting will be an issue:
- Currently we're using some std::error_code to indicate IR errors (no our own error code enums until we get this PR merged: https://github.com/y-scope/clp/pull/486). Using these error codes won't give users a clear error message on incomplete IR stream
- For incomplete IR stream, we should set the exception to `PyDeserializerBuffer::get_py_incomplete_stream_error()` which is a customized CPython exception specific to incomplete stream error.

haiqi96: Fair enough, let's keep ur current version

haiqi96 (approved):
For title, how about

feat: Add Python class Deserializer to deserialize CLP key-value pair IR stream into log events.

LinZhihao-723:
> For title, how about
> 
> feat: Add Python class Deserializer to deserialize CLP key-value pair IR stream into log events.

As discussed offline, we decided to stick to the original PR title since the phrasing is smoother and more readable

PR #95, opened 2024-11-26
refactor: Fix clang-tidy warnings in `Py_utils`.

### LINKED ISSUE #96: Fix clang-tidy warnings for all C++ source files
### Request

Until 0c66d04a43badd77ed51e6d029b07de94bbcb02c, we have clang-tidy added to the linting tasks but disabled due to unresolved clang-tidy warnings across all C++ source files. 
We should gradually resolve all these warnings.

### Possible implementation

We shall send a serious of PRs to fix the warnings. Before all warnings are fixed, we should modify the lint task to only execute clang-tidy against fixed files (as discussed [here](https://github.com/y-scope/clp-ffi-py/pull/95#discussion_r1859140486)). Eventually, we should have warnings resolved across all files, and we can set the lint task to run on all files (back to the config in 0c66d04a43badd77ed51e6d029b07de94bbcb02c)

### New linting changes introduced to fix linting issues:
- bugprone-suspicious-stringview-data-usage: #95 
- Missing includes for third-party dependencies:
  - Having a directory `wrapped_dependency_headers` in parallel to `clp_ffi_py`
  - Each dependency has a wrapped header similar to `Python.hpp` introduced in #81:
    - Exporting the super header
    - Exporting any internal headers used by `clp_ffi_py` when `CLP_FFI_PY_ENABLE_LINTING` is defined
- Disable cppcoreguidelines-pro-type-vararg: #101 
- Add clang-tidy command line option to bypass problematic lines in msgpack source code as a temporary solution for #109: #108.

LinZhihao-723:
# Description

This is one of PRs to resolve #96 

This PR fixes clang-tidy warnings in `Py_utils` under clang-tidy 19 using the latest clang-tidy config from https://github.com/y-scope/yscope-dev-utils/blob/main/lint-configs/.clang-tidy
One noticeable change is that we introduced a `consteval` function to wrap `constexpr` string view to resolve this warning: https://clang.llvm.org/extra/clang-tidy/checks/bugprone/suspicious-stringview-data-usage.html. We might also need this for other CLP repos

# Validation performed

- Ensure workflow passed
- Ensure unit tests all passed locally
- Ensure clang-tidy warnings all resolved for `Py_utils`

## Summary by CodeRabbit

- **New Features**
	- Introduced new string handling functions for improved efficiency, including `get_c_str_from_constexpr_string_view` and `parse_py_string_as_string_view`.

- **Improvements**
	- Updated function name constants to enhance clarity and type safety.
	- Changed header file inclusions for better functionality and compatibility.

- **Configuration**
	- Updated linting task configurations for enhanced readability and adjusted paths for linting tools.
	- Introduced a configuration file for Clang-Tidy to customize linting behaviour.


src/clp_ffi_py/Py_utils.cpp
```diff
@@ -2,24 +2,29 @@
...
-
+constexpr std::string_view cPyFuncNameGetFormattedTimestamp{"get_formatted_timestamp"};
+constexpr std::string_view cPyFuncNameGetTimezoneFromTimezoneId{"get_timezone_from_timezone_id"};
 constexpr std::string_view cPyFuncNameSerializeDictToMsgpack{"serialize_dict_to_msgpack"};
-PyObjectStaticPtr<PyObject> Py_func_serialize_dict_to_msgpack{nullptr};
-
 constexpr std::string_view cPyFuncNameParseJsonStr{"parse_json_str"};
-PyObjectStaticPtr<PyObject> Py_func_parse_json_str{nullptr};
+
+// NOLINTBEGIN(cppcoreguidelines-avoid-non-const-global-variables)
```
LinZhihao-723, line 21: Since this is quite common in Python ffi repo, can we disable this rule only for this repo?
If so, shall we add a clang-tidy config file inside `src/clp_ffi_py` directory?

kirkrodrigues: * If we disable the check globally, we don't need to clutter up the code with the disabling comments;
* but at the same time, we could unintentionally create a non-const global variable that should've been const.
* Of course, if you just put the non-const variable in the disabled section, then the effect is the same;
* but I feel like having the check disabled explicitly in the code will remind us that we're doing something special.
* Adding the check in each file doesn't seem like that much extra effort.

So overall, I feel like the current approach is slightly better.

src/clp_ffi_py/Py_utils.cpp
```diff
@@ -2,24 +2,29 @@
...
+PyObjectStaticPtr<PyObject> py_func_get_formatted_timestamp{nullptr};
```
kirkrodrigues: To stay consistent with our previous naming rules, can we add a clang-tidy override to include the `Py_xxx` pattern for variable names?

kirkrodrigues: I seem to get an error about no header providing PyObject.

src/clp_ffi_py/utils.hpp
```diff
@@ -85,6 +86,16 @@ auto handle_traceable_exception(clp::TraceableException& exception) noexcept ->
 template <typename T>
 [[maybe_unused]] constexpr bool cAlwaysFalse{false};
 
+/**
+ * A consteval to get the C-string from a constexpr string view.
```
kirkrodrigues:
```suggestion
```

kirkrodrigues (changes requested):

src/clp_ffi_py/Py_utils.cpp
LinZhihao-723: which Python version were you using? `PyObject` is defined in `object.h` and it's already included in our `Python.hpp`

src/clp_ffi_py/Py_utils.cpp
LinZhihao-723: Sure, let's keep it this way

src/clp_ffi_py/Py_utils.cpp
kirkrodrigues: Python 3.8, but I guess we don't support that for linting.

LinZhihao-723: Fixed by using `readability-identifier-naming.GlobalVariableIgnoredRegexp` to ignore any `Py_xxx`, added in the child directory (`src/clp_ffi_py`) clang-tidy config file

LinZhihao-723: yeah, I was doing linting in Python3.10

lint-tasks.yml
```diff
@@ -103,7 +103,9 @@ tasks:
     cmds:
       - task: ":utils:clang-tidy"
         vars:
-          FLAGS: "--config-file={{.ROOT_DIR}}/.clang-tidy -p {{.CLP_FFI_PY_COMPILE_COMMANDS_DB}}"
+          FLAGS: >
+            --config-file={{.CLP_FFI_PY_CPP_SRC_DIR}}/.clang-tidy
+            -p {{.CLP_FFI_PY_COMPILE_COMMANDS_DB}}
```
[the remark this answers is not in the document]
LinZhihao-723: It's added in this PR.

src/clp_ffi_py/.clang-tidy
```diff
@@ -0,0 +1,6 @@
+InheritParentConfig: true
+
+CheckOptions:
+  # Variable naming rules
+  # NOTE: we set this to allow PyObject global variables to start with `Py_`
```
kirkrodrigues:
```suggestion
  # Allow PyObject global variables to start with `Py_`
```

src/clp_ffi_py/.clang-tidy
```diff
@@ -0,0 +1,6 @@
+InheritParentConfig: true
+
+CheckOptions:
+  # Variable naming rules
+  # NOTE: we set this to allow PyObject global variables to start with `Py_`
+  readability-identifier-naming.GlobalVariableIgnoredRegexp: "Py_.*"
```
kirkrodrigues:
```suggestion
  readability-identifier-naming.GlobalVariableIgnoredRegexp: "Py(_[a-z0-9]+)+"
```

The current regex is not restrictive enough.

lint-tasks.yml
kirkrodrigues:
```suggestion
          FLAGS: >-
            --config-file "{{.CLP_FFI_PY_CPP_SRC_DIR}}/.clang-tidy"
            -p "{{.CLP_FFI_PY_COMPILE_COMMANDS_DB}}"
```

* As our [guidelines](https://docs.yscope.com/dev-guide/contrib-guides-taskfiles.html#paths-in-commands) mention, paths in commands should be quoted.
* Instead of `=` for one arg and space-separation for the other, let's use one consistent style.

lint-tasks.yml
```diff
@@ -103,7 +103,9 @@ tasks:
...
           SRC_PATHS:
             ref: ".G_CPP_LINT_DIRS"
```
kirkrodrigues, line 110: How about (you suggested this a long time ago) setting this to only the files in which we've fixed all the clang-tidy issues? That way, we have a record of which files we've finished and we can enable automatic checking to ensure we're not breaking fixed files in future PRs.

kirkrodrigues (changes requested):

lint-tasks.yml
LinZhihao-723: kk, sorry I didn't realize there's a guideline for task files. Will read it soon

lint-tasks.yml
LinZhihao-723: How about:
- Creating an issue tracking the clang-tidy warnings
- Start a new PR after this one merged to keep track of files already fixed, and enable checking

kirkrodrigues: Sure

LinZhihao-723: #96

kirkrodrigues (approved):

PR #127, opened 2025-01-20
feat: Add support for serializing/deserializing auto-generated key-value pairs in the new IR format.

LinZhihao-723:
# Description

This PR updates to the latest CLP core commit to adapt the feature enhancement introduced in [this](https://github.com/y-scope/clp/pull/653) PR. From this PR, we have the following supports:
- Add support for serializing auto-generated kv pairs using `Serializer`'s API
- Update `Deserializer`'s implementation to adapt the latest core ffi API changes
- Add support for retrieving both auto-generated and user-generated kv pairs from `KeyValuePairLogEvent`'s API
- Add support for creating a `KeyValuePairLogEvent` object from auto-gen and user-gen Python dictionaries

To test the new features, this PR:
- Remain the old serder tests by serializing empty auto-generated kv pairs (testing for backward compatibility)
- Improve the serder tests to serialize a nested Python dictionary as auto-generated kv pairs (testing for new features)

# Validation performed

- Ensure the doc string site can be successfully rendered with the expected content.
- Run unit tests with a debug version of Python with address sanitizer, ensuring that there's no illegal memory access or memory leaks caused by any code inside the ffi library.

## Release Notes

- **New Features**
	- Enhanced log event handling with separate auto-generated and user-generated key-value pairs.
	- Introduced a new function for unpacking msgpack maps, improving msgpack processing capabilities.

- **Bug Fixes**
	- Updated msgpack map processing to handle different types of key-value pairs more robustly.

- **Refactor**
	- Restructured internal methods to support more granular log event management.
	- Improved error handling and type checking for serialization and deserialization.

- **Tests**
	- Updated tests to validate the new structure of log event serialization and deserialization.

These updates provide more flexible and precise log event processing, allowing for better separation of automatically generated and user-generated metadata.


clp_ffi_py/ir/native.pyi
```diff
@@ -86,8 +86,8 @@ class FourByteDeserializer:
     ) -> Optional[LogEvent]: ...
 
 class KeyValuePairLogEvent:
-    def __init__(self, dictionary: Dict[Any, Any]): ...
-    def to_dict(self) -> Dict[Any, Any]: ...
+    def __init__(self, auto_gen_kv_pairs: Dict[Any, Any], user_gen_kv_pairs: Dict[Any, Any]): ...
```
junhaoliao, line 89: Shall we swap the order of `user_gen_kv_pairs` & `auto_gen_kv_pairs` and make `auto_gen_kv_pairs` optional? That way we can achieve backward compatibility.

(That said, I can see any generic developers might not need to call this constructor.)

clp_ffi_py/ir/native.pyi
```diff
@@ -86,8 +86,8 @@ class FourByteDeserializer:
...
+    def to_dict(self) -> Tuple[Dict[Any, Any], Dict[Any, Any]]: ...
```
junhaoliao, line 90: That whether kv-pairs are auto-generated might not matter to all developers. Can we also provide a helper to return a single dict?

clp_ffi_py/ir/native.pyi
LinZhihao-723: This constructor is mainly for testing purpose; we usually construct `KeyValuePairLogEvent` directly from the deserializer in CPython level.
But I think your comment might make sense in `Serializer`'s API changes.

clp_ffi_py/ir/native.pyi
LinZhihao-723: hmmm, my understanding is this should be the low level API, so we should return whatever given on the serialization end.
If you are suggesting merging two dicts into one, I think it would also be problematic because:
- We can't simply merge them into one since there might be key conflicts in two dicts
- If we add keys on the top level, like `{"auto_gen": auto_gen_dict, "user_gen": user_gen_dict}`, it would be a little confusing since the returned dictionaries aren't exactly what user inputs are and we need more documents to clarify things. It might be better just to return two dictionaries separately.

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.hpp
```diff
@@ -85,11 +85,14 @@ class PyKeyValuePairLogEvent {
...
-     * @return A new reference to the created dictionary on success.
+     * Converts the underlying key-value pair log event into Python dictionaries.
+     * @return A new reference to a Python tuple containing a pair of Python dictionaries on
+     * success:
+     * - A Python dictionary for auto-generated key-value pairs.
+     * - A Python dictionary for user-generated key-value pairs.
      * @return nullptr on failure with the relevant Python exception and error set.
      */
-    [[nodiscard]] auto to_dict() -> PyDictObject*;
+    [[nodiscard]] auto to_dict() -> PyObject*;
```
junhaoliao, line 95: Just curious, any reason we don't return `PyTupleObject*` instead?

junhaoliao (commented):
I checked the implementation and docs and didn't find anything too odd. I left some suggestions regarding the Python interfaces. Feel free to object.

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.hpp
LinZhihao-723: I used `Py_BuildValue` to create the tuple, which returns `PyObject*`.
Reference: https://docs.python.org/3/c-api/arg.html#c.Py_BuildValue

LinZhihao-723: Tuple APIs would also return `PyObject*` directly, like this one (which is equivalent to `Py_BuildValue`): https://docs.python.org/3/c-api/tuple.html#c.PyTuple_Pack

clp_ffi_py/ir/native.pyi
junhaoliao: Right, i was proposing that we can simply merge the two dicts into one.  That there can be potential key conflicts, though rare, is valid concern.

If we are to keep returning a tuple, having the user-kv-pairs as the first element might be more flexible to developers. e.g., they can write
```
kv_pairs, = log_event.to_dict()
```
to only get the user inserted pairs, if they don't care about the auto generated ones.

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.hpp
junhaoliao: Gotcha. Right, I read from the https://docs.python.org/3/c-api/arg.html#c.Py_BuildValue docs that the API is not guaranteed to return a tuple, but since we have the format string fixed, it should not return anything else i assume.

clp_ffi_py/ir/native.pyi
LinZhihao-723: Thought a little more. I feel like we shouldn't allow optional input even in `Serializer`. As this is the low level API, both values should be explicitly given. Also, most users won't use this API directly anyways. They should use Python logging library to log their kv pairs, which will be handled as user-generated by the loglib's code.

clp_ffi_py/ir/native.pyi
LinZhihao-723: hmmm, the current ordering strictly follows the underlying ffi core APIs.
If really necessary, we can add APIs like `get_user_gen_kv_pairs_as_dict` and `get_auto_gen_kv_pairs_as_dict` in future PRs?

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.hpp
LinZhihao-723: Swith to [PyTuple_Pack](https://docs.python.org/3/c-api/tuple.html#c.PyTuple_Pack) since it doesn't need argument parsing.

junhaoliao (approved):
The Python interfaces lgtm. Code logic matches their descriptions. Preferably someone with more C++ experience should also take a look. Otherwise all is good.

clp_ffi_py/ir/native.pyi
```diff
@@ -86,8 +86,8 @@ class FourByteDeserializer:
     ) -> Optional[LogEvent]: ...
 
 class KeyValuePairLogEvent:
```
junhaoliao, line 88: I'm not too sure how we are going to implement the compatibility layer for CLP Text IR Streams. In the future, will log events also be accessed through this interface?

LinZhihao-723: Yes, `timestamp` will be stored inside the auto-generated dictionary and `msg` will be stored inside the user-generated dictionary.

PR #131, opened 2025-01-23
feat(KeyValuePairLogEvent): Extend `to_dict()` to support custom encoding and error handling for `UnicodeDecodeError` (resolves #129).

### LINKED ISSUE #129: Add support to bypass invalid UTF8 byte sequence errors when converting `KeyValuePairLogEvent` to dictionaries.
### Request

As explained in [this](https://github.com/y-scope/clp/issues/686) issue from CLP core, converting `KeyValuePairLogEvent` that contains invalid UTF8 strings to Python dictionaries will fail due to a Unicode error. We should give users a solution to bypass this error.

### Possible implementation

- Instead of using `Unicode_FromStringAndSize` [here](https://github.com/y-scope/clp-ffi-py/blob/main/src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp#L599), we can use [PyUnicode_decode](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_Decode)
- In `KeyValuePairLogEvent.to_dict`, we add two arguments to receive `encoding` and `error` similar to [str](https://docs.python.org/3/library/stdtypes.html#str). We should default the values to "utf-8" and "strict", respectively.

This will allow users to specify an approach to bypass the invalid UTF8 bytes, such as `error="replace"`

LinZhihao-723:
# Description

This PR implements #129.
In this PR, we extend the current `KeyValueLogEvent.to_dict` to receive standard Python unicode encoding types and error handlers from users, which helps decode strings with invalid UTF-8 sequences. The method mirrors Python's [str](https://docs.python.org/3/library/stdtypes.html#str) object with [encoding](https://docs.python.org/3/library/codecs.html#standard-encodings) and [errors](https://docs.python.org/3/library/codecs.html#error-handlers). The underlying CPython implementation will use [PyUnicode_decode](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_Decode) to apply the given encoding standard and error handler.

However, we realize that if we replace the current `PyUnicode_FromStringAndSize` with `PyUnicode_Decode`, the runtime of the private micro-benchmark will increase by around 7-8% to deserialize an IR stream. This is because `PyUnicode_Decode` method will involve more underlying string compare operations on the given `errors` and `encoding` to determine which decoding algorithm to use. This check is needed for every `PyUnicode_Decode` call when converting a C++ string to a Python Unicode object.

To solve this issue, we templated `to_dict`'s CPython implementation to receive different string-to-py-unicode methods. This allows us to check `encoding` and `errors` at the beginning and decide whether we can use `PyUnicode_FromStringAndSize` to avoid the overhead of `PyUnicode_Decode` when the giving arguments are the default. This helps to reduce the runtime regression if users don't specify any errors or encoding types different to the default ones.

Due to the fact that we need to template `to_dict`'s C++ implementation, we move some methods from the source file to the header, which include:
- `PyKeyValuePairLogEvent::to_dict` (template function)
- `serialize_node_id_value_pair_to_py_dict` (template function, used by `PyKeyValuePairLogEvent::to_dict`)
- `insert_kv_pair_into_py_dict` (template function, used by `serialize_node_id_value_pair_to_py_dict`)
- `PyDictSerializationIterator` (class, used by `serialize_node_id_value_pair_to_py_dict`)

Notice that `PyKeyValuePairLogEvent_internal` namespace is used for internal helpers, as discussed in our latest coding guideline [here](https://docs.yscope.com/dev-guide/contrib-guides-cpp.html#internal-linkage).

# Validation performed

- Ensure the API doc can be successfully compiled and rendered.
- Add new unit test case to:
  - Serialize/deserialize IR that contains non-UTF8 strings
  - Ensure a log event with such a string:
    - Can't be converted to Python dictionary with the default encoding and errors setting
    - Can't be converted to Python dictionary with the wrong encoding type
    - Can be successfully converted to Python dictionary with the proper encoding type (cp932)
    - Can be successfully converted to Python dictionary with `"ignore"` error handler

## Release Notes

- **New Features**
  - Enhanced `to_dict()` method with customizable string encoding options.
  - Added support for specifying encoding and error handling when converting log events to dictionaries.

- **Improvements**
  - Improved flexibility in handling Unicode string conversions.
  - More robust error management during dictionary serialization.

- **Testing**
  - Added comprehensive test coverage for invalid UTF-8 encoding scenarios.

These updates provide greater control over string encoding when working with log events, allowing users to specify custom encoding and error-handling strategies.


src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
```diff
@@ -263,15 +109,29 @@ PyKeyValuePairLogEvent_init(PyKeyValuePairLogEvent* self, PyObject* args, PyObje
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays)
 PyDoc_STRVAR(
         cPyKeyValuePairLogEventToDictDoc,
-        "to_dict(self)\n"
+        "to_dict(self, encoding=\"utf-8\", errors=\"strict\")\n"
         "--\n\n"
         "Converts the log event into Python dictionaries.\n\n"
+        "The `encoding` and `errors` parameters allow customization of Unicode behaviors, similar"
+        " to those in Python's `str` objects, for converting C strings into Python Unicode"
+        " objects.\n\n"
```
junhaoliao: how about
```suggestion
        " to those in Python's `bytes.decode()` method, for converting C strings into Python"
        " Unicode objects.\n\n"
```

or
```suggestion
        " to those in Python's `codecs.decode()` method, for converting C strings into Python"
        " Unicode objects.\n\n"
```

tests/test_ir/test_key_value_pair_log_event.py
```diff
@@ -41,3 +42,45 @@ def test_basic(self) -> None:
...
             num_files_tested, 0, f"No test files found in directory: {test_data_dir}"
         )
+
+    def test_invalid_utf8_encoding(self) -> None:
+        """
+        Tests handling of invalid UTF-8 encoded strings.
+        """
+        encoding_type: str = "cp932"
+
+        # msgpack map: {"key": 0x970x57}, where "0x970x57" is encoded using "cp932"
```
junhaoliao: do you mean
```suggestion
        # msgpack map: {"key": 0x970x5c}, where "0x970x5c" is encoded using "cp932"
```

junhaoliao (changes requested):
the logic and Python interfaces lgtm. let's fix some nits in the docs

junhaoliao (commented):
For the PR title, how about
```
feat(KeyValuePairLogEvent): Extend `to_dict()` to support custom encoding and error handling (resolves #129).
```

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
LinZhihao-723: I referred to `str` since `PyUnicode_Decode` does so: https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_Decode

LinZhihao-723:
> For the PR title, how about
> 
> ```
> feat(KeyValuePairLogEvent): Extend `to_dict()` to support custom encoding and error handling (resolves #129).
> ```

Add "for Unicode" at the end to clarify what is the encoding target

src/clp_ffi_py/ir/native/PyKeyValuePairLogEvent.cpp
junhaoliao: Gotcha. For better clarity, can we say "Python's `str()` built-in function" instead of "Python's `str` objects"?

LinZhihao-723: kk make sense

LinZhihao-723: Applied some rewording to make it as close to `PyUnicode_Decode`'s documentation as possible.

junhaoliao:
> > For the PR title, how about
> > ```
> > feat(KeyValuePairLogEvent): Extend `to_dict()` to support custom encoding and error handling (resolves #129).
> > ```
> 
> Add "for Unicode" at the end to clarify what is the encoding target

<strike>
It seems Python 3 raises `UnicodeError` or its subclasses whenever some encoding / decoding error happens regardless whether the encoding is "Unicode" or not. e.g.,
```
PyDev console: starting.
Python 3.9.13 (tags/v3.9.13:6de2ca5, May 17 2022, 16:36:42) [MSC v.1929 64 bit (AMD64)] on win32
"我爱学习".encode("gb2312").decode("ascii")
Traceback (most recent call last):
  File "C:\Users\Junhao\AppData\Local\Programs\PyCharm Professional 2\plugins\python-ce\helpers\pydev\pydevconsole.py", line 364, in runcode
    coro = func()
  File "<input>", line 1, in <module>
UnicodeDecodeError: 'ascii' codec can't decode byte 0xce in position 0: ordinal not in range(128)
```

It's definitely Python being confusing in the first place. 
</strike>

I had some misconceptions about how Python uses Unicode as the encoding for all `str` objects.
Still, I think adding `UnicodeError` / `UnicodeDecodeError` might promote better clarity since we are providing ways to mitigate errors that happen during "bytes in arbitrary encoding" -> "Unicode".
```
feat(KeyValuePairLogEvent): Extend `to_dict()` to support custom encoding and error handling for `UnicodeDecodeError` (resolves #129).
```

junhaoliao (approved):

LinZhihao-723:
> > > For the PR title, how about
> > > ```
> > > feat(KeyValuePairLogEvent): Extend `to_dict()` to support custom encoding and error handling (resolves #129).
> > > ```
> > 
> > 
> > Add "for Unicode" at the end to clarify what is the encoding target
> 
> It seems Python 3 raises `UnicodeError` or its subclasses whenever some encoding / decoding error happens regardless whether the encoding is "Unicode" or not. e.g., ``` PyDev console: starting. Python 3.9.13 (tags/v3.9.13:6de2ca5, May 17 2022, 16:36:42) [MSC v.1929 64 bit (AMD64)] on win32 "我爱学习".encode("gb2312").decode("ascii") Traceback (most recent call last): File "C:\Users\Junhao\AppData\Local\Programs\PyCharm Professional 2\plugins\python-ce\helpers\pydev\pydevconsole.py", line 364, in runcode coro = func() File "", line 1, in UnicodeDecodeError: 'ascii' codec can't decode byte 0xce in position 0: ordinal not in range(128) ```
> It's definitely Python being confusing in the first place.
> 
> I had some misconceptions about how Python uses Unicode as the encoding for all `str` objects. Still, I think adding `UnicodeError` / `UnicodeDecodeError` might promote better clarity since we are providing ways to mitigate errors that happen during "bytes in arbitrary encoding" -> "Unicode".
> 
> ```
> feat(KeyValuePairLogEvent): Extend `to_dict()` to support custom encoding and error handling for `UnicodeDecodeError` (resolves #129).
> ```

Yeah, in Python3 `str` and unicode object are the same. So `decode(encoding=xxx)` is basically saying we decode a byte sequence using xxx encoding convention, and turn the results into a Unicode object.

PR #132, opened 2025-01-23
chore: Update to the latest clp submodule commit.

LinZhihao-723:
# Description

This PR updates to the latest CLP submodule commit to pull the following changes in core ffil:
- The fix in https://github.com/y-scope/clp/pull/687.
- The feature improvement in https://github.com/y-scope/clp/pull/691.

# Validation performed

- Ensure all workflows passed.

## Summary by CodeRabbit

- **Chores**
	- Updated subproject reference to a new commit identifier.

PR #133, opened 2025-01-26
feat: Add support for serializing/deserializing user-defined stream-level metadata.

LinZhihao-723:
# Description

This PR depends on #132, #134.
This PR adds support for serializing/deserializing user-defined stream-level metadata. As discussed offline, we decided to constrain the metadata input into a JSON string to simplify the type-checking.

# Validation performed

- Update existing unit tests to ensure user-defined metadata can be successfully serialized/deserialized.

## Summary by CodeRabbit

- **New Features**
	- Added support for user-defined metadata in serialization and deserialization processes.
	- Introduced ability to pass and retrieve custom metadata during stream processing.
	- New function for serializing Python dictionaries to JSON strings.

- **Documentation**
	- Updated method signatures and documentation to explain new metadata functionality.

- **Tests**
	- Added test cases to validate user-defined metadata handling.
	- Implemented error checking for invalid metadata inputs.

The changes enhance the serialization framework by allowing users to attach and retrieve custom metadata during stream operations, providing more flexibility in data processing.


src/clp_ffi_py/ir/native/PyDeserializer.cpp
```diff
@@ -67,6 +72,23 @@ PyDoc_STRVAR(
...
+ */
+PyDoc_STRVAR(
+        cPyDeserializerGetUserDefinedMetadataDoc,
+        "get_user_defined_metadata(self)\n"
+        "--\n\n"
+        "Gets the user-defined stream-level metadata.\n\n"
+        ":return:\n"
+        "    - The deserialized user-defined stream-level metadata, loaded as a"
+        " dictionary.\n"
+        "    - None if user-defined stream-level metadata was not given in the deserialized"
```
junhaoliao, line 86: Would there be less handling for API users if we return an **empty dictionary** if "user-defined stream-level metadata was not given in the deserialized IR stream"?

src/clp_ffi_py/ir/native/PyDeserializer.cpp
```diff
@@ -153,6 +180,48 @@ CLP_FFI_PY_METHOD auto PyDeserializer_deserialize_log_event(PyDeserializer* self
...
+CLP_FFI_PY_METHOD auto PyDeserializer_get_user_defined_metadata(PyDeserializer* self) -> PyObject* {
+    auto const* user_defined_metadata{self->get_user_defined_metadata()};
+    if (nullptr == user_defined_metadata) {
+        Py_RETURN_NONE;
+    }
+
+    if (false == user_defined_metadata->is_object()) {
+        PyErr_SetString(
+                PyExc_TypeError,
+                "User-defined stream-level metadata is not a JSON object."
```
junhaoliao: If the type of this JSON value is not an object, it means either the file is not created using a Serializer we provide, or something went wrong in our code, right?

Though i imagine such cases are rare, would it be more helpful to debug such case if we include the JSON value type in the error message?

src/clp_ffi_py/ir/native/PyDeserializer.cpp
```diff
@@ -153,6 +180,48 @@ CLP_FFI_PY_METHOD auto PyDeserializer_deserialize_log_event(PyDeserializer* self
...
+        PyErr_Format(
+                PyExc_RuntimeError,
+                "Failed to serialize the user-defined stream-level metadata into a JSON string."
+                " Error: %s",
+                ex.what()
+        );
+        return nullptr;
+    }
+
+    PyObjectPtr<PyObject> py_metadata_dict{py_utils_parse_json_str(json_str)};
```
junhaoliao, line 202: sorry for not being able find any previous discussions on this - we have considered the https://github.com/pybind/pybind11_json and other alternatives discussed at https://www.github.com/nlohmann/json/issues/1014 , right?

just curious, for more efficient conversions, do we have plans to make use of those or create our own bindings?

src/clp_ffi_py/ir/native/PyDeserializer.cpp
junhaoliao: Though, it does seem in the Serializer we do permit users to provide a `None` or "{}" value. I haven't traced down to the CLP code but I think the given metadata string will be encoded as-is?

src/clp_ffi_py/ir/native/PySerializer.cpp
```diff
@@ -42,6 +43,9 @@ PyDoc_STRVAR(
         ":param buffer_size_limit: The maximum amount of serialized data to buffer before flushing"
         " it to `output_stream`. Defaults to 64 KiB.\n"
         ":type buffer_size_limit: int\n"
+        ":param user_defined_metadata: JSON-encoded string containing user-defined stream-level"
```
junhaoliao: I think an easier to use interface is to accept a dict type. If we're time-constraint to implement that, that's fine.

src/clp_ffi_py/ir/native/PySerializer.cpp
```diff
@@ -276,7 +284,39 @@ CLP_FFI_PY_METHOD auto PySerializer_init(PySerializer* self, PyObject* args, PyO
...
+            PyErr_Format(
+                    PyExc_RuntimeError,
+                    "Failed to parse `%s`: %s",
+                    static_cast<char const*>(keyword_user_defined_metadata),
+                    json_str_data
+            );
+            return -1;
+        }
+        optional_user_defined_metadata = std::move(parsed_user_defined_metadata);
+    }
```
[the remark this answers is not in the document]
junhaoliao: I agree with the rabbit if we're to accept a `str` type metadata.

src/clp_ffi_py/ir/native/PyDeserializer.cpp
LinZhihao-723: Since this isn't in any of the critical path anyways, I would probably defer it to future release

src/clp_ffi_py/ir/native/PyDeserializer.cpp
LinZhihao-723: Yeah, the only possible case is that someone created a stream that follows our format but doesn't enforce the use of a JSON object as the metadata. Technically we should check this in the ffi core. Let me think about it...

src/clp_ffi_py/ir/native/PyDeserializer.cpp
LinZhihao-723: I would prefer to return `None` indicating that the user-defined metadata wasn't specified in the source.
It is possible to encode an empty dictionary as the metadata though.

src/clp_ffi_py/ir/native/PySerializer.cpp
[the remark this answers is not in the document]
LinZhihao-723: This type validation will be handled by ffi core API, when we use this parsed object to create a serializer.

src/clp_ffi_py/ir/native/PySerializer.cpp
LinZhihao-723: That's true. I think we can fix it with a Python API binding.

LinZhihao-723: Done.

clp_ffi_py/utils.py
```diff
@@ -48,6 +48,19 @@ def serialize_dict_to_msgpack(dictionary: Dict[Any, Any]) -> bytes:
...
+    Serializes the given dictionary into a JSON string.
+
+    :param dictionary: The dictionary to serialize.
+    :return: JSON string of serialized dictionary.
+    :raises: TypeError The given input is not a dictionary.
+    """
+    if not isinstance(dictionary, dict):
+        raise TypeError("The type of the input object must be a dictionary.")
+    return json.dumps(dictionary)
+
```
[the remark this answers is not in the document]
LinZhihao-723: Is re-throwing the same exception here making any difference?

src/clp_ffi_py/ir/native/PyDeserializer.cpp
LinZhihao-723: As improved in https://github.com/y-scope/clp/pull/695, we moved this check to the ffi core. The latest commit should have this check removed (however, we need to merge #134 first).

clp_ffi_py/utils.py
```diff
@@ -48,6 +48,19 @@ def serialize_dict_to_msgpack(dictionary: Dict[Any, Any]) -> bytes:
     return msgpack.packb(dictionary)
 
 
+def serialize_dict_to_json_str(dictionary: Dict[Any, Any]) -> str:
+    """
+    Serializes the given dictionary into a JSON string.
+
+    :param dictionary: The dictionary to serialize.
+    :return: JSON string of serialized dictionary.
```
junhaoliao:
```suggestion
    :return: JSON string of the serialized dictionary.
```

junhaoliao (dismissed):

junhaoliao (commented):
the pr title is fine to be used as the final commit message

junhaoliao (approved):

